PR #4, opened 2016-05-15
Serialize errors to show stack trace

mujx:
I added a module to serialize the errors from the workers.


josdejong:
Thanks, that sounds good, I will review your PR asap.

josdejong:
Thanks @mujx , nice solution!

josdejong:
Your solution is now available in v1.2.0

PR #7, opened 2016-09-12
Fixes Worker bug in PhantomJS

natlibfi-arlehiko:
Workaround for invalid implementation of Worker in PhantomJS/QtWebkit (See https://github.com/ariya/phantomjs/issues/14534)


josdejong:
Thanks for this fix @natlibfi-arlehiko!

josdejong:
I've just released version 1.3.1 containing your fix.

PR #8, opened 2016-09-14
Reimplement Error serialization

natlibfi-arlehiko:
Replace converting Error-objecting using serializerr to custom implementation. The conversion done by serializerr allows functions in the object but Web Workers cannot handle functions when transferring objects (See: https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm  _Error and Function objects cannot be duplicated by the structured clone algorithm; attempting to do so will throw a DATA_CLONE_ERR exception._)

This conversion implementation loses the prototype object which means that e.g. 'TypeError' will become just 'Error' in the main code.


josdejong:
Thanks for this improvement and simplification @natlibfi-arlehiko. I think it's acceptable to lose the prototype object and return a generic `Error` instance.

josdejong:
I've released `v2.0.0` containing your fix, thanks!

PR #10, opened 2016-09-19
Add minWorkers option

sergei202:
Add `minWorkers` option to ensure a minimum number of workers always up and running.

Setting `minWorkers` to true will create `numCPUs-1` workers.

If `minWorkers` is greater than `maxWorkers`, `maxWorkers` will be set equal to it.


josdejong:
Thanks for your PR Sergei, looks good!

I think there is one edge case that should be handled too: when workers crash and are terminated, we should immediately spawn new workers to keep the minimum required number of workers available.

sergei202:
That's a good point, I didn't think of that edge case.  I'll try handling that night or tomorrow, along with some documentation in `README.md` that I forgot to include.

It would also be nice to have a way for the workers to notify the pool when they are ready.  Maybe we can add an `onReady` callback (or Promise) option for when all the workers are initialized (along with some way for them notify that they are ready).  What are your thoughts on that?

josdejong:
Good point about the docs! Thanks, take your time.

An `onReady` callback could be useful, though I really want to keep the API concise and focused so there should be a good use case for it. Do you have actual use case for this?

sergei202:
Hey @josdejong!

I put the code to respawn crashed workers in `Pool.js`.
I looked at putting it in `WorkerHandler.js` but I'd have to pass an extra argument to the constructor and it seemed clunky.

Let me know what you think!

josdejong:
Thanks @sergei202 I think `Pool.js` is indeed the right place to handle initializing new workers after a crash. A few small remarks:

- I've just merged a PR from @natlibfi-arlehiko which offers a property `environment.cpus` which can now be used straight away instead of determining the number of CPU's in `Pool.js`.
- Can you replace tabs in the code with 2 space indentation?
- Now we have two places in the code with a for loop which initializes enough workers to meet `minWorkers`. I think it would be neat to move this logic into a method `ensureMinWorkers()` and use that.
- I was thinking, the shortcut to set `minWorkers: true` is very convenient but it's probably not clear to users what "true" means. How about changing it to something like `minWorkers: 'max'`? (or do you have other suggestions?)

sergei202:
Hey @josdejong, those are great suggestions.  I'll implement them later this week.

josdejong:
Cool, looking forward to your updates.

sergei202:
@josdejong: Implemented the items discussed and a few other things:
- Clean up tabs / indents
- Use `environment.cpus`
- Change `minWorkers:true` to `minWorkers:'max'`
- Created a `ensureMinWorkers()` function
    - I created it on the `Pool` prototype so it has access to `this`. Since it's a private function, do you want an underscore in front of it, or even create it as a standalone function and pass the `Pool` reference?
- Created unit tests for `minWorkers` options.  I didn't get around to adding a test for crashed workers, but can if you think it's important.
- Changed `minWorkers` option to accept `0`, which will simply disable `minWorkers`.

Let me know what you think!

josdejong:
Thanks @sergei202 , looks good! I'll merge your PR now.

josdejong:
I've released v2.1.0 containing the new option `minWorkers` :)

PR #13, opened 2016-10-03
Export environment properties

natlibfi-arlehiko:
Export the following properties about the environment for convenience:

  - platform - The Javascript platform. Either 'node' or 'browser'
  - is_main_thread - Whether the code is running in the main thread or not (Workers)
  - cpus - The number of CPUs/cores available


natlibfi-arlehiko:
Commit message needs to be fixed

PR #15, opened 2016-10-03
Worker ready

natlibfi-arlehiko:
Implement worker 'ready' message  & export environment


lib/WorkerHandler.js
```diff
@@ -116,6 +129,12 @@ function WorkerHandler(script) {
     me.processing = {};
   }
 
+  // send all queued request to worker
+  function dispatchQueuedRequests()
+  {
+    me.request_queue = me.request_queue.filter(me.worker.send.bind(me.worker));
```
josdejong: Shouldn't this be `forEach` instead of `filter`?

```js
me.request_queue.forEach(me.worker.send.bind(me.worker));
me.request_queue = [];
```

lib/WorkerHandler.js
```diff
@@ -80,26 +80,39 @@ function WorkerHandler(script) {
...
-      }
-      else {
-        task.resolver.resolve(response.result);
+  // The ready message is only sent if the worker.add method is called (And the default script is not used)
+  if (!script) {
+    this.worker.ready = true;
+  }
+
+  // queue for requests that are received before the worker is ready
+  this.request_queue = [];
```
josdejong: Can you please rename `request_queue` to `requestQueue`? Or maybe `taskQueue` or something? At least lowerCamelCase :).

lib/WorkerHandler.js
natlibfi-arlehiko: This is more functional approach, but since it's not used elsewhere so much I'll change it to #forEach.

natlibfi-arlehiko:
The problem with not committing `dist` is that when you end up using branches as dependencis you need to have the transpiled file available.

One option would be to transpile at npm's `postinstall` step, but I understand this not a recommended solution according to NPM people.

I think including the `dist` in a branch separate from the PR one would solve this.

lib/WorkerHandler.js
josdejong: Methods like `forEach`, `map`, and `filter` are all functional. I understand what you did and why but the way you used `filter` is isn't really wat it is intended for: sending messages inside a `filter` action is a side effect which is the opposite of a functional approach, and these send messages don't return anything so it's misleading to filter based on a function which always returns `undefined`. Anyway, thanks for changing it!

josdejong:
I don't like committing these `dist` folders. github should be for source code and not for built artifacts. 

The main reason is support for bower, not npm. Bower doesn't allow a build step and is basically gives you a dump of the source code from a github project. To make a library useful via bower you have to commit generated bundles in your source code. 

With npm you can publish generated bundles rather than just source code, and we don't have this problem there. 

What I normally do with my libraries is just commit the source code, and when I publish a new version of a library, generate the bundles under `/dist` and commit them once. It works. An alternative could be to reserve the master branch for the sake of bower, and put the generated bundles there, and do source code in other branches.

PR #16, opened 2016-10-04
Implement worker 'ready' message

natlibfi-arlehiko:
Implement worker 'ready' that enables asynchronous initialization of workers (For using e.g. AMD modules)


josdejong:
Thanks a lot @natlibfi-arlehiko , looks good!

Issue #19, opened 2017-02-13
Node child forks can't use the debugger


romainbessugesmeusy:
Hi Jos,

It looks like I can't use node debugger while spawning child processes, because each worker try to use the same port. Example:

```
Debugger listening on 127.0.0.1:51884
Error: listen EADDRINUSE 127.0.0.1:51884
    at Object.exports._errnoException (util.js:1023:11)
    at exports._exceptionWithHostPort (util.js:1046:20)
    at Agent.Server._listen2 (net.js:1261:14)
    at listen (net.js:1297:10)
    at doListening (net.js:1396:7)
    at _combinedTickCallback (internal/process/next_tick.js:77:11)
    at process._tickCallback (internal/process/next_tick.js:98:9)
```

It seems that child forks have to be created with the `--debugger={port}`. My first guess would be to change this line (WorkerHandler.js:79):

```javascript
    this.worker = node.require('child_process').fork(this.script);
```

What do you think?

Thanks again for your work, cheers from France

romainbessugesmeusy:
I think it's interesting to mention this discussion on the IntelliJ / WebStorm board

https://intellij-support.jetbrains.com/hc/en-us/community/posts/206193339-Debugging-node-child-processes

josdejong:
Thanks for reporting Romain!

> It seems that child forks have to be created with the --debugger={port}

Wouldn't that always run the childs in debug mode, also if you just want to run in production?

romainbessugesmeusy:
Yeah I fear you're right. Furthermore I haven't found yet how to attach a remote debugger to child process inside Webstorm... Maybe this could be passed when creating the pool, as a secondary argument.

```javascript
var workerpool = require('workerpool');

// create a worker pool using an external worker script
var pool = workerpool.pool(__dirname + '/myWorker.js', {
   node: {
       childProcessExecArgs: (childIndex) => return '--debugger=' + (53027 + childIndex)
   }
});
 ```

josdejong:
hm that's odd. I can hardly believe we're the first stumbling upon this problem.

Maybe we could check whether the `--debugger` option is in the arguments of the main process, and only then add it in the `childProcessExecArgs` with a unique port. Or is that what you mean?

From the [thead you mentioned](https://intellij-support.jetbrains.com/hc/en-us/community/posts/206193339-Debugging-node-child-processes?page=1#community_comment_206895729) I understand that it should be enough to pass an argument `--debug-brk` which will let node.js automatically pick a free port.

[...]

PR #21, opened 2017-04-06
Add node --inspect support for child processes

### LINKED ISSUE #2: passing Parameters to workers
It would be cool if you could pass parameters to the workers.
For example beeing able to increase the stack-size would be nice.
Also always passing the parent parameters is a problem, if you are using an IDE like VS Code, because it uses a debug port, which is blocked by the main process, when the worker try to open it (so the workers crash).

tptee:
Also adds support for passing `args` and `opts` to `fork()`.

Tested with `--inspect` and `--debug-brk`, all seems well.

Resolves #2 and #19

### DIFF
--- a/lib/Pool.js
+++ b/lib/Pool.js
@@ -20,6 +20,12 @@ function Pool(script, options) {
   this.workers = [];  // queue with all workers
   this.tasks = [];    // queue with tasks awaiting execution
 
+  options = options || {};
+
+  this.forkArgs = options.forkArgs || [];
+  this.forkOpts = options.forkOpts || {};
+  this.debugPortStart = options.debugPortStart || 43210;
+
   // configuration
   if (options && 'maxWorkers' in options) {
     if (!isNumber(options.maxWorkers) || !isInteger(options.maxWorkers) || options.maxWorkers < 1) {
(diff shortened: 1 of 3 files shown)


josdejong:
Thanks Tyler, I will review your PR asap, hopefully within a few days

lib/Pool.js
```diff
@@ -274,12 +284,38 @@ Pool.prototype.stats = function () {
...
     }
   }
 };
 
 /**
+ * Ensure that the maxWorkers option is a positive integer
+ * @param {*} value
+ * @returns {boolean} returns true when value is a positive integer
+ */
+function validateMaxWorkers(maxWorkers) {
```
josdejong: Good idea to extract `validateMaxWorkers` and `validateMinWorkers` in a separate function, only you forget to actually use them :D

lib/WorkerHandler.js
```diff
@@ -6,6 +6,10 @@ var environment = require('./environment');
 // used to prevent webpack from resolving requires on node libs
 var node = {require: require};
 
+var assign = environment.platform === 'node'
```
josdejong: I think this can be simplified to 

```js
var assign = require('object-assign');
```

since this library can be used in both node.js and the browser. That may also prevent some weird situations in the future if some day we want to use `assign` for other purposes as well but in the browser it's set to null.

josdejong:
@tptee I suppose you're occupied these days, I'll merge your solution now and fix the comments I made myself. Thanks again!

josdejong:
I've just released `v2.2.1` containing this fix.

PR #29, opened 2017-08-18
Fixed debug issue - look for '--inspect' within argument strings, instead of exact match

jimsugg:
Recent versions of node allow extended forms of the --inspect parameter, such as --inspect-brk or --inspect=<port>. (See https://nodejs.org/en/docs/inspector/) This change will enable the inspector on a  worker process if any parameter contains '--inspect', rather than requiring an exact match. I needed to add this in order to be able to use breakpoints in the worker process in WebStorm.

### DIFF
--- a/lib/WorkerHandler.js
+++ b/lib/WorkerHandler.js
@@ -34,8 +34,10 @@ function resolveForkOptions(opts) {
 
   const execArgv = [];
 
-  const inspectorActive = process.execArgv
-    .indexOf('--inspect') !== -1;
+  var inspectorActive = false;
+  process.execArgv.some(function(value) {
+    if (value.indexOf('--inspect') !== -1) { inspectorActive = true; return true; }
+  });
   const debugBrk = process.execArgv
     .indexOf('--debug-brk') !== -1;
 


josdejong:
Thanks @jimsugg , good improvement. 

I think you could simplify your code a bit:
```js
var inspectorActive = process.execArgv.some(function(value) {
   return value.indexOf('--inspect') !== -1;
});
```

There is one disadvantage of using `.some` though: we would have to drop the current support for IE10. We could do either that, or use a `for` loop instead of `Array.some`. What do you think?

jimsugg:
I think we should go with a for loop. I was trying for compatible code, but wasn't aware that IE10 had a problem with 'some'. Please feel free to optimize and improve compatibility.

Thanks for a great package!

If I can find time to work on an implementation to utilize streams for binary data transfer, I will submit that also. I thought it would be important for my project, but for my case I can work around it by using base64 encoding to serialize. That's only feasible in my project because the data isn't that big. For larger data, streams would definitely be preferable.

Jim

josdejong:
:+1:  I will adjust your code to use a for loop.

> If I can find time to work on an implementation to utilize streams for binary data transfer, I will submit that also.

That would be great!

josdejong:
I've refactored your PR (see https://github.com/josdejong/workerpool/commit/84a96b61d77754ba22f849bef1d5db233bbd2165), and released v2.2.4 containing your fix. 

Would be great if you could double check that the issue is indeed solved in v2.2.4.

jimsugg:
I tested with your latest code, and it’s working fine - thanks!

josdejong:
ok thanks!

PR #30, opened 2017-08-25
Added callback in WorkerHandler.terminate to prevent zombie processes

jimsugg:
Please take a look at this. This is not intended to be a final pull request proposal, but it is the primary work needed to solve a big problem I was having with zombie processes when testing some advanced scenarios using worker processes with fairly complex scripts. I was seeing instances where the main process never terminated properly due to this. I also wanted to add terminate functionality that would notify when it was complete (via a Promise.) This is helpful in shutdown scenarios.

In investigating the shutdown problems I was seeing, I noticed your TODO comment about using a callback to prevent zombies. So I implemented that. 

I also added terminateAndNotify functions to both WorkerHandler and Pool to return Promises to track termination of processes. I added these as separate functions with the idea of maintaining compatibility. On further thought, I think it may be reasonable just to fold this into the main terminate functions. I don't think there would be any bad side effects to returning a Promise from them, even if it isn't really used. Tomorrow I will try this out and see how it goes. Let me know if you have objections to this idea. If I get that working, I will check it in and you can take a look at that also.

While I was at it, I renamed Pool.close to Pool.terminate, as per your TODO comment. There is still a 'close' function which is an alias which is marked as deprecated. This keeps it from being a breaking change. If you are ok with this, I will change all the references to close as well.


jimsugg:
In the latest commit, I have simplified the Pool object by removing the separate terminateAndNotify method. The terminate method now returns a Promise that resolves when all processes have been terminated. A timeout can be specified, and if any process fails to terminate within the timeout period, the Promise is rejected with a TimeoutError. This is a release candidate for the pull request.

josdejong:
Thanks, that looks promising. I will review your PR soon but this is quite a busy weekend.

jimsugg:
Thanks - that's fine. For our project, we can pull it out of the fork until
you have a chance to check it out and merge it. Let me know if you see any
way to improve it or have any questions.

 The additional test I wrote was an attempt to test for a race condition I
initially encountered (and fixed,) but I could never get the test to
trigger the same situation that I saw when I ran with my much more complex
worker script. I left the additional test in anyway.

Jim

lib/WorkerHandler.js
```diff
@@ -289,11 +294,35 @@ WorkerHandler.prototype.terminate = function (force) {
     }
     this.terminating = false;
     this.terminated = true;
+    if (this.terminationHandler) {
+      this.terminationHandler(this);
```
josdejong: I don't fully understand how the (async?) callback is handled. As far as I can see the only place it's called is here, and this is called synchronously, which would make the callback redundant.

lib/Pool.js
```diff
@@ -264,16 +272,33 @@ Pool.prototype._removeWorker = function(worker) {
...
+  workers.forEach(function (worker) {
+    const termPromise = worker.terminateAndNotify(force, timeout)
+      .then(removeWorker);
+    promises.push(termPromise);
   });
+  return Promise.all(promises).then();
+};
 
-  this.workers = [];
+// DEPRECATED
```
josdejong: :+1:

lib/Pool.js
```diff
@@ -264,16 +272,33 @@ Pool.prototype._removeWorker = function(worker) {
...
+  const removeWorker = f.bind(this);
+
+  const promises = [];
+  const workers = this.workers.slice();
+  workers.forEach(function (worker) {
+    const termPromise = worker.terminateAndNotify(force, timeout)
+      .then(removeWorker);
+    promises.push(termPromise);
   });
+  return Promise.all(promises).then();
```
josdejong: Is there a special reason to append this line with `.then()` ? That doesn't do anything. (same on line 301)

lib/Pool.js
```diff
@@ -264,16 +272,33 @@ Pool.prototype._removeWorker = function(worker) {
...
-Pool.prototype.clear = function (force) {
-  this.workers.forEach(function (worker) {
-    // TODO: implement callbacks when a worker is actually terminated, only then clear the worker from our array
-    //       else we get zombie child processes :)
-    worker.terminate(force);
+Pool.prototype.terminate = function (force, timeout) {
+  var f = function (worker) {
+    this._removeWorkerFromList(worker);
+  };
+  const removeWorker = f.bind(this);
```
josdejong: I would prefer to use `var` only and not `const`, since we don't have a transpiler in place which will make the library work on older browsers too (IE 10).

lib/WorkerHandler.js
```diff
@@ -261,8 +262,9 @@ WorkerHandler.prototype.busy = function () {
  *                                  after finishing all tasks currently in
  *                                  progress. If true, the worker will be
  *                                  terminated immediately.
+ * @param {function} [callback=null] If provided, will be called when process terminates.
  */
-WorkerHandler.prototype.terminate = function (force) {
+WorkerHandler.prototype.terminate = function (force, callback) {
```
josdejong: Would it make sense to simply return a Promise instead of passing a callback? Most of the API is Promise based anyway :)

jimsugg: Yep, in theory, that sounds great. I tried it, and I couldn't make it work. The problem, as best as I could tell, has to do with the fact that WorkerHandler.terminate is called in the 'message' and 'error' handlers. So the scenario is this:

lib/WorkerHandler.js
jimsugg: 1. It is called, of course, only if one has been passed in to the original Workerhandler.terminate call.
2. It isn't redundant, because that part of the code is only executed if/when the process is not busy. If the process is busy, that code is not executed, and the 'terminating' flag is set. So if the process is busy, the callback is not immediately called.
3. In the busy case, the terminate function is called again when the process finally finishes - this is done by the 'on message' handler that is set up n the WorkerHandler constructor. In that second terminate call, the process will not be busy, and at that point the code above will be executed, and the callback is called at that point.
4. Now that I look at it to explain it, maybe we should also call the termination handler if the process errors out (in 'on error'.)

Let me know if it is still not clear.

lib/Pool.js
jimsugg: OK. I have been writing in Typescript mostly with tslint, which is very opinionated, and that has trained me to use const without thinking. var is fine.

lib/Pool.js
jimsugg: No, not really. It got rid of a warning in WebStorm that I wasn't handling the promise, which can of course just be ignored. The only argument I can make for it, which is somewhat weak, is that it clearly signals our intention to ignore the promise. I am fine with dropping that.
I decide to leave the 'clear' call in most of the tests because it really isn't necessary in the tests to change to using a promise. That little WebStorm warning showed up in all the tests without that, and it just annoyed me. 🙂

lib/WorkerHandler.js
jimsugg: And of course, if the process is not busy, the callback is called immediately, (and the Promise in TerminateAndNotify is resolved immediately,) which is what you want.

jimsugg: Bottom line is that it works in my complex case - without it the process would never shut down properly unless I put in an artificial timeout, because I had several child processes that were still busy. With this, everything terminates nicely. I was trying to replicate the situation in a simple unit test, but I haven't been able to do that so far.

jimsugg: I should maybe add that our scenario is node.js only - it's an Electron app - so all my testing is done there, and using node child processes. I have no way to test this in the browser case, but it seems it should be ok there as well.

jimsugg:
I'm thinking that in the onError function the WorkerHandler constructor, we should do this:
`  function onError(error) {
    me.terminated = true;
    if (me.terminating && me.terminationHandler) {
      me.terminationHandler(me);
    }
    me.terminating = false;

    for (var id in me.processing) {
      if (me.processing.hasOwnProperty(id)) {
        me.processing[id].resolver.reject(error);
      }
    }
    me.processing = {};
  }
`

jimsugg:
I'm thinking of adding this to onError in WorkerHandler constructor:
```
    if (me.terminating && me.terminationHandler) {
      me.terminationHandler(me);
    }
    me.terminating = false;

```

jimsugg:
Latest checkin addresses issues from most recent discussion. Also added better unit tests for the termination wait scenarios.

lib/WorkerHandler.js
josdejong: I think it should be possible to move the `var resolver = Promise.defer();` structure from `terminateAndNotify` to `terminate` and return a promise there. And both callback and Promise should resolve only once or course. It's not a big deal though, I'm ok with leaving this a callback structure (we can always do a refactoring later on if we want).

lib/WorkerHandler.js
josdejong: Thanks for the updates. I now see what I was missing: `WorkerHandler.terminate` is _first_ called via the `terminateAndNotify`. When not busy, it immeditately calls the callback function. When busy, `terminate` will be called _again_ after the task finished, but now _without_ a new callback function as argument, in which case the callback function (cached as `this.terminationHandler`) is called. 

Maybe the flow becomse a little bit more clear when moving setting of the `terminationHandler` from `terminate()` to `terminateAndNotify()`, since the callback is only passed when calling `terminateAndNotify`?

```js
  if (typeof callback === 'function') {
    this.terminationHandler = callback;
  }
```

Or give a promise based solution another try: on construction of `WorkerHandler`, create a promise
```js
this.terminated = Promise.defer();
```
this promise is returned by `terminateAndNotify`:
```js
WorkerHandler.prototype.terminateAndNotify = function (force, timeout) {
  // ...
  return this.terminated;
}
```
And resolve the promise as soon as the worker terminates or errors:
```js
this.terminated.resolve(this);
```

josdejong:
Now I understand your solution I think you've created a solid solution. 

It would be great to actually have a unit test that would generate a zombie process without your fix. Right now I can only take your word for it, and I totally believe you that this fixes the issue, but this way we're not protected against any regressions in the future. What do you think?

jimsugg:
I believe that the two unit tests that I added in the latest commit should
in fact test this. However, I have not gone back to the old code and tried
them with that to make sure they fail there. I am pretty sure they would,
but I will verify this.

Jim

josdejong:
Ok, that's good to hear. I will play around with the unit tests a bit more - see if I can get zombies with the old solution.

josdejong:
@jimsugg  I've just been playing around to see whether I can get zombie child_process but so far no success. Do you know of a simple example which causes a child_process not to be closed (in the old situation)?

josdejong:
Whoops I almost forgot that this PR is still open. Will merge it now (still no luck producing zombie processes, but the solution works for you...).

josdejong:
@jimsugg your fix for the zombie processes should be available now in `v2.3.0`, thanks again and sorry for the delay!

jimsugg:
Thanks. Sorry I didn’t get back to you on further tests - we are really
jammed on our project. I have thought that the complexity of our background
process may be a contributing factor, so it may be challenging to really
replicate in a test. I will keep it on my radar, though, because it would
be good to have a unit test for that. BTW, we may submit another PR for
some changes we needed to make it work in Electron. The test to determine
whether we are running under node or a in browser is tricky in that case.

josdejong:
yeah I can imagine that. Good luck with the struggles in your project!

PR #39, opened 2018-07-24
Fixed bug where tasks that are cancelled in a Pool's queue causes following tasks to not run

greemo:
This is a simple bugfix. Can you please commit this change to source and tests, and create a new npm bundle.

If you would like to add me as a maintainer, please go ahead.


josdejong:
Thanks a lot @greemo for fixing this issue! I've just released v2.3.1 containing your fix.

PR #41, opened 2018-09-06
Add license field to a package.json

greyd:
Our company uses license checker tools which breaks build if it cannot get a license for the package. I believe this change would help to resolve license info automatically.


josdejong:
Thanks @greyd !

josdejong:
The license field should now be available in `v2.3.2`

PR #42, opened 2018-09-12
remove space

sagotsky:
Our license checker chokes on the extra space before apache.  Mind fixing it?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "workerpool",
-  "license": " Apache-2.0",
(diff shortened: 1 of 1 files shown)


josdejong:
:+1: thanks. Would be nice as well if your license checker would be robust against this :D

josdejong:
Should be fixed now in `v2.3.3`

sagotsky:
Much appreciated.  I agree our checker should trim strings first.  At least
it let me whitelist " apache 2.0"

PR #49, opened 2019-01-18
Add support for Yarn PnP

Turbo87:
[Yarn Plug'n'Play](https://github.com/yarnpkg/rfcs/pull/101) is using a custom resolver to lookup dependencies from a machine-wide global location, instead of keeping `node_modules` duplicated in every project.

When `yarn run` is used and it detects that it is running in PnP mode it will call the child process using `--require .pnp.js`, with `.pnp.js` being a file generated per-project by Yarn. This will enable PnP mode for the child process and make the PnP API available via `require('pnpapi')`.

Since `workerpool` is also forking off child processes it would need a similar mechanism to ensure that those child processes are able to resolve their dependencies correctly. This PR adds the necessary `--require` argument to the fork option if PnP mode is detected by checking `process.versions.pnp`.

In case you're wondering: we're working on making [`ember-cli`](https://github.com/ember-cli/ember-cli) compatible with Yarn PnP and that is using [`broccoli-babel-transpiler`](https://github.com/babel/broccoli-babel-transpiler), which is using `workerpool` under the hood.

/cc @arcanis @rwjblue

### DIFF
--- a/lib/WorkerHandler.js
+++ b/lib/WorkerHandler.js
@@ -42,6 +42,13 @@ function resolveForkOptions(opts) {
     }
   }
 
+  // if the host process is running using Yarn PnP we should start the
+  // child processes in the same mode
+  if (process.versions.pnp) {
+    var pnpPath = require.resolve('pnpapi');
+    execArgv.push(`--require ${pnpPath}`);
+  }
+
   return assign({}, opts, {
     forkArgs: opts.forkArgs,
     forkOpts: assign({}, opts.forkOpts, {


rwjblue:
Chatted with @turbo87 in DM, I think we should move to inheriting `NODE_OPTIONS` instead (I’m somewhat surprised we aren’t already inheriting all of `process.env`).

This change would essentially be following the same move in Yarn itself (away from using `—require` to specifying `NODE_OPTIONS`) in https://github.com/yarnpkg/yarn/pull/6629.

rwjblue:
Ah, apparently `child_process.fork` does not automatically inherit the current processes `process.env` ( `child_process.spawn` has this behavior). That explains why it doesn’t “just work”. 

See here for the two implementations (on Node 10):

* spawn -
https://github.com/nodejs/node/blob/1ae0511b942c01c6e0adff98643d350a395bf101/lib/child_process.js#L398-L552
* fork -
https://github.com/nodejs/node/blob/1ae0511b942c01c6e0adff98643d350a395bf101/lib/child_process.js#L61-L114

So the question here is, should we inherit all of `process.env` or only `process.env.NODE_OPTIONS`?

Turbo87:
@rwjblue unfortunately, it's not quite as easy. `fork()` uses `spawn()` under the hood and that correctly forwards the existing env vars. but in our case when Yarn PnP mode is enabled via `ember` instead of using `yarn run` it will not have a `NODE_OPTIONS` env var with the correct value.

Turbo87:
we will most likely resolve this another way (adjusting `NODE_OPTIONS` ourselves once we detect Yarn PnP mode) so I'll close this PR for now. sorry for the noise 😉

josdejong:
:+1: thanks anyway Tobias for taking the time to put together a PR :)

PR #50, opened 2019-01-26
[FEATURE] [Fixes #38] Use Node 11’s WorkerThreads when available

### LINKED ISSUE #38: Node just got worker_threads
Hey, we have been using workerpool for some time to help parallelize ember-cli builds. Thanks for a great project!

Anyways, I wanted to share that node just added experimental support for worker_threads: https://github.com/nodejs/node/pull/20876 And this most likely could be of benefit to this project.

I'll likely kick the tires a bit, if it turns out good would their be interest in a PR providing support for those who opt-in?

stefanpenner:
This seems to work, some notes:

* Under heavy load, WorkerThreads crash less then the old subprocess fork model
* Able to saturate CPU without subprocesses
* No more leaking sub-processes
* ~~added `WORKERPOOL_FORCE_SPAWN` environment variable, to force spawn. I found this useful when testing...~~
* added `options.nodeWorker = 'auto' | 'thread' | 'process'`

image


stefanpenner:
Early perf numbers suggest slightly worse throughput, I'll investigate....

stefanpenner:
While profiling the above, I noticed some minor hot-spots, so I fixed them.

stefanpenner:
The remaining overhead in this library seems to be GC caused by the custom promise implementation, if it matters some future work could eliminate that.

dist/workerpool.js
```diff
@@ -978,7 +978,7 @@ return /******/ (function(modules) { // webpackBootstrap
 	        me.processing[id].resolver.reject(error);
 	      }
 	    }
-	    me.processing = {};
+	    me.processing = Object.create(null);
```
stefanpenner: `me.processing` is ulltimately a growing/shrinking dictionary object, using `Object.create(null)` both ensures the prototype isn't polluted, and also gives a good hint to the JIT. This gets most JITs to skip trying to figure out what shape the object is, and rather go straight to dictionary mode.

lib/Pool.js
```diff
@@ -187,9 +189,7 @@ Pool.prototype._next = function () {
       if (task.resolver.promise.pending) {
         // send the request to the worker
         var promise = worker.exec(task.method, task.params, task.resolver)
-          .then(function () {
-            me._next(); // trigger next task in the queue
-          })
+          .then(me._boundNext)
```
stefanpenner: I noticed this callback was causing a decent amount of extra GC, so now we always reuse the same callback instance.

lib/Pool.js
```diff
@@ -223,21 +223,22 @@ Pool.prototype._next = function () {
  */
 Pool.prototype._getWorker = function() {
   // find a non-busy worker
-  for (var i = 0, ii = this.workers.length; i < ii; i++) {
-    var worker = this.workers[i];
-    if (!worker.busy()) {
+  var workers = this.workers;
+  for (var i = 0; i < workers.length; i++) {
```
stefanpenner: Memoizing the array length ended up being counter productive, as I believe the internal bounds check was required anyways... Removing this removed a hot-spot the profiler raised.

lib/WorkerHandler.js
```diff
@@ -127,12 +158,12 @@ function WorkerHandler(script, options) {
       // find the task from the processing queue, and run the tasks callback
       var id = response.id;
       var task = me.processing[id];
-      if (task) {
+      if (task !== undefined) {
```
stefanpenner: Coercion here seemed to pop up in the profiler, so I simply made the check explicit.

lib/WorkerHandler.js
```diff
@@ -127,12 +158,12 @@ function WorkerHandler(script, options) {
...
         // remove the task from the queue
         delete me.processing[id];
 
         // test if we need to terminate
-        if (me.terminating) {
+        if (me.terminating === true) {
```
stefanpenner: Coercion here seemed to pop up in the profiler, so I simply made the check explicit.

lib/WorkerHandler.js
```diff
@@ -157,11 +188,11 @@ function WorkerHandler(script, options) {
     me.terminating = false;
 
     for (var id in me.processing) {
-      if (me.processing.hasOwnProperty(id)) {
+      if (me.processing[id] !== undefined) {
```
stefanpenner: `me.processing` is an `Object.create(null)` so both does not need `hasOwnProperty` and does not support `hasOwnProperty`. So it can simply become `!== undefined`.

simonihmig:
Gave this a try with https://github.com/stefanpenner/broccoli-concat-analyser.

With `workerpool@3.0.0`: **224s** (average across 3 runs)
With this branch: **208s**
(both running node v11.8.0)

The offloaded work falls more into the "Long tasks" category, so the performance improvements are nice, but not exceptional. I once placed with splitting up the worker tasks into smaller units of work in that library, but that didn't yield any performance improvements back then. Could try that again now...

josdejong:
Wow, that's not only implementing WorkerThreads, but also a bunch of optimizations :sunglasses:  thanks @stefanpenner! And thanks for explaining the reason behind the optimizations :+1:

I have a few feedback points:

1. Can you have a look at the indentation of the `var WorkerThreads; try { ... } catch { ... }` in `WorkerHandler.js`?
2. Is the `process.env.WORKERPOOL_FORCE_SPAWN` meant as an option to specify whether to use child processes or worker threads? Maybe we should turn this into an option that you can pass when creating a workerpool, something `workerpool.pool('/my/worker/script.js', { forceUseChildProcess: true }`?
3. I thought the most important reason for worker threads is not performance improvement but a reduction of memory. So I did a small experiment: I changed the your version of the fibonacci example to pass option `{ maxWorkers: 500 }`, changed the number of loops to at least 1000, and removed `pool.terminate()` so I can see the memory usage in htop. A rough test showed that using processes memory usage is 3.3 GB and with threads 2.5 GB. I had expected a more spectacular difference.
4. I would love to get rid of the custom Promise implementation. Most important reason I did need it in the first place is to have a way to cancel and timeout promises, and I wanted to keep the bundle small. But let's keep that as a separate topic.

josdejong:
Forgot to say: I have been testing your PR, it works like a charm!

stefanpenner:
> 1. Can you have a look at the indentation of the var WorkerThreads; try { ... } catch { ... } in WorkerHandler.js?

done

> Is the process.env.WORKERPOOL_FORCE_SPAWN meant as an option to specify whether to use child processes or worker threads? Maybe we should turn this into an option that you can pass when creating a workerpool, something workerpool.pool('/my/worker/script.js', { forceUseChildProcess: true }?

For my use-case (many worker pools, deeply used in build tooling) the `env` var was more ergonomic.  But I see no reason for a flag, as you suggested, to exist.

That being said, I think we need some design on what those flags are and how they operate.
One idea:

* add opt-in/opt-out and "best available" flags
* by default, WorkerThreads are be opt-in for next minor version
* by default, WorkerThreads are be opt-out for next major version


> I thought the most important reason for worker threads is not performance improvement but a reduction of memory. So I did a small experiment: I changed the your version of the fibonacci example to pass option { maxWorkers: 500 }, changed the number of loops to at least 1000, and removed pool.terminate() so I can see the memory usage in htop. A rough test showed that using processes memory usage is 3.3 GB and with threads 2.5 GB. I had expected a more spectacular difference.

Ya, workers aren't really light-weight, they are still more or less "share nothing", at-least when it comes to the primary consumers memory.

stefanpenner:
> I would love to get rid of the custom Promise implementation. Most important reason I did need it in the first place is to have a way to cancel and timeout promises, and I wanted to keep the bundle small. But let's keep that as a separate topic.

Ya, we can keep that as a separate discussion. Reading the promise implementation + usage I noticed some issues, but decided to only fix those affecting this PR.

As, for Promise cancellation, that is a pretty contentious topic, if we want to have that chat, yes it should be a separate issue :)

stefanpenner:
It's worth noting, that `process.chdir()` isn't available from a `WorkerThread`


```js
const {
  Worker, isMainThread
} = require('worker_threads');

if (isMainThread) {
  const worker = new Worker(__filename, {
    stdout: false,
    stderr: false,
  });
} else {
  process.chdir('/');
}
```

stefanpenner:
List of differences (from the official doc)

```
Notable differences inside a Worker environment are:

The process.stdin, process.stdout and process.stderr may be redirected by the parent thread.
The require('worker_threads').isMainThread property is set to false.
The require('worker_threads').parentPort message port is available,
process.exit() does not stop the whole program, just the single thread, and process.abort() is not available.
process.chdir() and process methods that set group or user ids are not available.
process.env is a read-only reference to the environment variables.
process.title cannot be modified.
Signals will not be delivered through process.on('...').
Execution may stop at any point as a result of worker.terminate() being invoked.
IPC channels from parent processes are not accessible.
The trace_events module is not supported.
Currently, the following differences also exist until they are addressed:

The inspector module is not available yet.
Native addons are not supported yet.
```

stefanpenner:
@josdejong I agree.

With one refinement, the following given no support for WorkerThreads, should fail early with an exception:
```js
nodeWorkers: 'thread'
```

josdejong:
> given no support for WorkerThreads, should fail early with an exception

ah, yes, very good point!

lib/WorkerHandler.js
```diff
@@ -25,6 +40,52 @@ function getDefaultWorker() {
   }
 }
 
+function setupBrowserWorker(script) {
```
stefanpenner: TODO: unit test

lib/WorkerHandler.js
```diff
@@ -25,6 +40,52 @@ function getDefaultWorker() {
...
+      callback(message.data);
+    });
+  };
+  worker.send = function (message) {
+    this.postMessage(message);
+  };
+  return worker;
+}
+
+function setupWorkerThreadWorker(script) {
```
stefanpenner: TODO: unit test

lib/WorkerHandler.js
```diff
@@ -25,6 +40,52 @@ function getDefaultWorker() {
...
+  };
+
+  worker.disconnect = function() {
+    this.terminate();
+  };
+
+  return worker;
+}
+
+function setupProcessWorker(script, options) {
```
stefanpenner: TODO: unit test

lib/WorkerHandler.js
```diff
@@ -1,9 +1,24 @@
+'use strict';
+
 var Promise = require('./Promise');
 var assign = require('object-assign');
 
 // determine environment
 var environment = require('./environment');
 
+function try_require(moduleName) {
```
stefanpenner: TODO: unit test

lib/WorkerHandler.js
```diff
@@ -74,40 +135,36 @@ function objectToError (obj) {
...
-    // on node.js, create a child process
-    forkOptions = resolveForkOptions(options);
-
-    this.worker = require('child_process').fork(
-      this.script,
-      forkOptions.forkArgs,
-      forkOptions.forkOpts
-    );
+    this.worker = setupBrowserWorker(this.script);
+  } else {
```
stefanpenner: TODO: unit test

lib/worker.js
```diff
@@ -19,8 +19,28 @@ if (typeof self !== 'undefined' && typeof postMessage === 'function' && typeof a
 }
 else if (typeof process !== 'undefined') {
   // node.js
-  worker.on = process.on.bind(process);
-  worker.send = process.send.bind(process);
+
+  var WorkerThreads;
```
stefanpenner: TODO: test

josdejong:
Looks really neat! I love how you neatly split creation of the three types of workers with three setup functions.

josdejong:
@stefanpenner would be great if you can find some time to finish this PR, I guess it's almost ready to merge?

stefanpenner:
Oops thanks for reminding me. I think I’m just missing an acceptance test, stressing actual WorkerThread work. I’ll do that shortly.

Anything else catch your eye?

josdejong:
> Anything else catch your eye?

All looks good to me. I'm really glad you spend effort in adding unit tests for the worker threads!

stefanpenner:
@josdejong sorry for the delay, I finally had time to flesh out the tests and refactor to ensure the error messages are thrown early enough.  Let me know if that looks good, I'll gladly make changes if needed.

josdejong:
Awesome! Will do another testing round within a few days.

stefanpenner:
Nice. Ya let me know if anything is sideways. I’ll do the same.

josdejong:
All looks good, thanks again @stefanpenner !

I've removed the `yarn.lock` file, I'm not using yarn so I'm quite sure it would be outdated soon doing more harm than good.

josdejong:
:tada: worker_threads are now available in `v3.1.0`

josdejong:
@stefanpenner I've just published `workerpool@5.0.0` which uses `worker_threads` by default instead of `child_process`. Would be great if you can give it a try.

PR #52, opened 2019-02-24
[BUGFIX] nodeWorker: auto now can actually choose WorkerThreads

stefanpenner:
Oops, caught my mistake when trying to actually use workerpool \w `nodeWorker: 'auto'`.
Added tests, to validate.

### DIFF
--- a/lib/WorkerHandler.js
+++ b/lib/WorkerHandler.js
@@ -53,6 +53,7 @@ function setupBrowserWorker(script, Worker) {
   // create the web worker
   var worker = new Worker(script);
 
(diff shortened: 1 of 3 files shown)


josdejong:
ow, that's odd! We've done quite some testing with it. I will do another test round tomorrow and publish your fix then (assuming all is ok).

josdejong:
All looks good :+1: . This is a bug with a clear and simple cause and fix :)

josdejong:
@stefanpenner your fix is now available in `workerpool@3.1.1`. Thanks!

PR #54, opened 2019-03-03
tests refactored for readability

izelnakri:


josdejong:
Thanks for the rewrite @izelnakri, looks way better! I would love to rewrite the code itself too ES6 too, but right now there is not yet a build tool in place to convert to ES5 before publishing. Having to write the source code in ES5 and tests in ES6 will be confusing, so I think it's essential that we first set up webpack (or another transpiler) before merging this PR. That will be a great step on itself. 

Would you be interested into getting webpack in place in this project Izel?

izelnakri:
Hi @josdejong, Im willing to maintain this library, very likely. My plan is to refactor majority parts of the source code with typescript -> transpile to es5 for node and browser and then cut a major release. We can also introduce a separate qunit test suite for browsers and perhaps ava to modernize the test suite later. Thanks for the library again.

josdejong:
:smiley: I wasn't looking directly for a new maintainer to take over the project though any contribution is very welcome.

We can consider a rewrite to TypeScript, though we should be careful not to make too big steps at a time. It could make sense though in that case to setup TypeScript, allow js files, and gradually migrate the code to TypeScript.

Issue #32, opened 2017-11-04
Better worker termination


astanciu:
It would be nice to have more advanced worker termination options, at least for node.
- The timeout is a false one. It just causes the promise to reject. I would expect the timeout to actually force kill
- Currently terminating with `worker.kill()`. If one wanted to implement some cleanup only way would be to trap SIGTERM. Would be nicer to support a method like `{onTermination: myCleanup}`, then have the timeout actually kill the process with `worker.kill('SIGKILL')`

josdejong:
Thanks, agree with both your points Alex.

If anyone is interested in addressing these points pleas let me know, help would be very welcome!

mramato:
@josdejong I assume this has never been implemented?  I took a look and couldn't figure out how to perform one-time cleanup on worker shutdown.

Implementing SIGTERM, as the original description suggests, creates a subtle race condition because `WorkerHandler.terminate` is synchronous and it has no way of knowing when the kill message it sent to the child process is complete and the process is actually shut down.

What I think I would propose is that if a user adds a `terminate` method to their worker, `WorkerHandler.terminate` would exec it and wait for it to resolve/reject before calling `worker.kill`.  We would have to bubble up this promise to `pool.terminate` which could now guarantee that all workers had a chance to shutdown cleanly before it resolves.

I did an initial pass on the implementation of this and it looks easy enough, the main problem is that everywhere else in the code treats `WorkerHandler.terminate` as synchronous and I wanted to run my proposal by you before spending the time to do the update (plus on the off chance that you have a better idea).

What do you think?

josdejong:
Ai! 

My first thought was that `.terminate()` didn't return a Promise, but looking at the code I saw that a Promise was actually returned and I assumed this was implemented in recent PR's.

It would be very nice if we improve `terminate`, it's tricky that we don't really have control about knowing when all workers are _actually_ killed. Would be great if you could work out a solution for this!

I had a short look at the node.js API for killing `child_process`. It looks like after calling `worker.kill()` we can wait for the a `close` and/or `exit` event which will let us know that the process is actually killed I think? At least we will end up with an async API. That may require some rework in the code indeed.

josdejong:
Help improving the `terminate` method would be very welcome! Anyone?

josdejong:
This should be fixed now in `v6.0.2` thanks to #176

PR #57, opened 2019-03-10
Update documentation and examples for pool.terminate.

### LINKED ISSUE #56: Is pool.terminate async?
The documentation treats `pool.terminate` as synchronous, but in practice (at least under Node.js) it returns a Promise.

It makes sense to me that it would be asynchronous, but I wanted to confirm before updating my own code.

If it is async, I'd be happy to open a PR to make sure the doc matches.

Thanks for an awesome library!

mramato:
`pool.terminate` returns a Promise, but the documentation and examples treated it as if it were synchronous.

Fixes #56

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ pool.exec(add, [3, 4])
       console.error(err);
     })
     .then(function () {
-      pool.terminate(); // terminate all workers when done
+      return pool.terminate(); // terminate all workers when done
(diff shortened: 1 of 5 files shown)


josdejong:
Before merging this PR, I think we first have to actually implement the async termination, we discovered that's not yet the case :roll_eyes: , see #32

PR #58, opened 2019-03-11
Improve worker exited unexpectedly error

stefanpenner:
When a worker exits unexpectedly, it can be tricky to figure out why.
To improve debugging, I've added some additional information to the existing error message:

* the status code or signal code
* that workerpools script
* spawnargs
* spawnfile
* stdout
* stderr
 
example output:

```js
Error: Workerpool Worker terminated Unexpectedly
    exitCode: `12`
    workerpool.script: `/Users/spenner/src/josdejong/workerpool/lib/worker.js`
    spawnArgs: `/usr/local/Cellar/node/11.10.0/bin/node,--inspect=43210,/Users/spenner/src/josdejong/workerpool/lib/worker.js`
    spawnfile: `/usr/local/Cellar/node/11.10.0/bin/node`
    stdout: `null`
    stderr: `null`

    at ChildProcess.<anonymous> (/Users/spenner/src/josdejong/workerpool/lib/WorkerHandler.js:263:13)
    at ChildProcess.emit (events.js:197:13)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:254:12)
```

PR #60, opened 2019-03-12
[Fixes #58] Allocate Debug ports safely, so multiple pools don’t acci…

stefanpenner:
…dentally overlap

### DIFF
--- a/lib/Pool.js
+++ b/lib/Pool.js
@@ -1,7 +1,8 @@
 var Promise = require('./Promise');
 var WorkerHandler = require('./WorkerHandler');
 var environment = require('./environment');
-
+var DebugPortAllocator = require('./debug-port-allocator');
+var DEBUG_PORT_ALLOCATOR = new DebugPortAllocator();
 /**
  * A pool to manage workers
  * @param {String} [script]   Optional worker script
@@ -24,7 +25,7 @@ function Pool(script, options) {
 
   this.forkArgs = options.forkArgs || [];
   this.forkOpts = options.forkOpts || {};
-  this.debugPortStart = options.debugPortStart || 43210;
+  this.debugPortStart = (options.debugPortStart || 43210);
   this.nodeWorker = options.nodeWorker;
 
   // configuration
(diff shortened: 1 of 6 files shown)


josdejong:
Wow, didn't think about that but indeed both pools try to use the same port.0

 I do like your approach of a central place to get a free port number, that makes sense. 

It may interesting too think about not keeping state in the `DebugPortAllocator` and find a free port by simply trying a range, though I'm not sure about the performance but when this is just about debugging that could be perfectly acceptable. Keeping state and having to release a port afterwards is more fragile than a dumb "just-try-until-you-find-something" approach. There are npm packages that can do this for us, like [get-port](https://www.npmjs.com/package/). How do you think about that @stefanpenner ?

stefanpenner:
In ember-cli, we use https://www.npmjs.com/package/portfinder and we have contributed many fixes to that (turns out to be a harder problem than expected).

I don't mind adding that, I avoided doing as I wasn't sure if you wanted it as a dependency.

josdejong:
> turns out to be a harder problem than expected

o really? Well, I don't think we should over-engineer this and add a large dependency just for this. In that case your current solution may be best and simplest after all.

stefanpenner:
@josdejong I think its "ok" for now. How about if this problem in-practice arises for us we add portFinder?

The issues we had with portFinder are largely addressed, the tricky bits arose due to:

* mixing ipv4 and ipv6 interfaces
* some programs binding to 4 not 6 others to 6 not 4
* some programs, such as the MacBook Pro touch-bar running on a port, but leaving that port open...

stefanpenner:
@josdejong I'm going to land this as-is, because the inability to debug builds is causing some grief.

I will though, gladly make any follow up changes :) Just let me know.

stefanpenner:
updated `history.md` https://github.com/josdejong/workerpool/commit/9793e3e4596668a1f37fdd4ef87b811000f87edb

stefanpenner:
Ah, I don't have NPM publish yet. Guess I can't actually ship this yet.

josdejong:
> I think its "ok" for now. How about if this problem in-practice arises for us we add portFinder?

Yes I agree, no need to over think this right now.

I was also thinking, this functionality doesn't need to be bundled for use in the browser since it's only needed in node.js, maybe we can do something smart there.

josdejong:
@stefanpenner shall I publish `v3.1.2` now?

josdejong:
Will do a publish now @stefanpenner.

stefanpenner:
@josdejong that would be wonderful :) Thank you kindly good sir.

josdejong:
You're welcome :) I published v3.1.2 almost an hour ago, you should be able to use it.

PR #70, opened 2019-06-19
copy --max-old-space-size argv to fork options

patte:
my workers died with `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`. This change copies over `--max-old-space-size=` to forkOpts.execArgv if set in `process.execArgv`.

Screenshot from 2019-06-20 01-22-26

### DIFF
--- a/lib/WorkerHandler.js
+++ b/lib/WorkerHandler.js
@@ -117,6 +117,12 @@ function resolveForkOptions(opts) {
     }
   }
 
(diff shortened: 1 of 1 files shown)


josdejong:
Nice, makes sense. Thanks for your PR Patrick. 

I can imagine that you would like to be able to set a different value of `--max-old-space-size` for your main process vs your workers, but let's start simple.

Issue #43, opened 2018-12-05
(Question) - how to import workerpool in es6 module


GOVYANSONG:
Our environment is typescript, es6 module and webpack. When I tried to compile my project, i received following webpack errors:

WARNING in ./node_modules/workerpool/lib/environment.js
    2:21-28 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
     @ ./node_modules/workerpool/lib/environment.js
     @ ./node_modules/workerpool/index.js
     @ ./ClientApp/util/Fetcher.ts
     @ ./ClientApp/util/index.tsx
     @ ./ClientApp/boot-client.tsx
    
    WARNING in ./node_modules/workerpool/lib/WorkerHandler.js
    8:21-28 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
     @ ./node_modules/workerpool/lib/WorkerHandler.js
     @ ./node_modules/workerpool/lib/Pool.js
     @ ./node_modules/workerpool/index.js
     @ ./ClientApp/util/Fetcher.ts
     @ ./ClientApp/util/index.tsx
     @ ./ClientApp/boot-client.tsx

I guess my question is how to import workerpool in es6 module. or is it possible.

josdejong:
The [`environment.js`](https://github.com/josdejong/workerpool/blob/master/lib/environment.js) file contains some commonjs tricks to prevent node.js specific code being bundled in the browser bundle. I think there is a refactoring required to make this work with ES6, we probably have to create two separate entry points (one for browser, one for node.js) or something like that. Anyone interested into looking into this?

I don't know if there are ways to exclude the library from TypeScript checking somehow. Webpack itself is capable of compiling the code so there may be a way to get this working as is.

GOVYANSONG:
Jos, thanks for the timely response. Very appreciated. After experimented with few twits in webpack in vain, we quickly explored other projects. Here is the one we are investigating: threads.js. It supports es6 and backward compatible besides other desirable features.

knnhcn:
I am getting a similar warning in Angular 6:

```
WARNING in ./node_modules/workerpool/lib/environment.js
2:21-28 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

WARNING in ./node_modules/workerpool/lib/WorkerHandler.js
8:21-28 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
Compiled with warnings.
```

Holgrabus:
[...]

PR #74, opened 2019-07-16
Allow setup platform in global environment

cawa-93:


josdejong:
Thanks for your PR Alex. What is your usecase exactly? Shouldn't we simply improve the browser vs node.js detection so that it's not needed to provide an environment variable?

Some thoughts: If we go for the environment variable

- it will be a good idea to make it unique so that it cannot conflict with other libraries, like `WORKERPOOL_PLATFORM` or so.
- We should document this feature

cawa-93:
I'll try to expand the browser's password. In fact, the execution environment is a browser. However, I am running a project assembly using a roller. And as far as I understand the library determines the execution medium as a node.

My code:
https://github.com/cawa-93/play-shikimori-online/blob/be1587393bcebb9993994d906ebc83313412cbe7/src/player/index.js#L12-L14

When I try to build a project I get the following error:
```
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
worker_threads (imported by node_modules/workerpool/lib/worker.js, worker_threads?commonjs-external)
os (imported by node_modules/workerpool/lib/environment.js, os?commonjs-external)
child_process (imported by node_modules/workerpool/lib/WorkerHandler.js, child_process?commonjs-external)
```

I thought that it would be great to be able to independently specify the execution environment that I need. No matter what the library calculates.

josdejong:
ah, that's interesting. It sounds very similar to #43, which still gives issues. This would suggest though that the cause may be wrong detection of node.js environment. That needs further investigation.

josdejong:
What do you think @cawa-93? Does it make sense to look into whether the environment is wrongly detected?

cawa-93:
If you plan to create several separate builds for the server and the browser, then this PR no sense. After all, I can just import the build I need regardless of the environment.
Otherwise, I think it's still necessary.
Automatic environment detection is great. However, you should allow developers the ability to manage this parameter, as you cannot anticipate all possible usage scenarios.

josdejong:
Yes that makes sense. Let's try to solve it that way.

josdejong:
Alex, I've just released `workerpool@4.0.0` with a changed bundle which should work out of the box in browser, node.js, and Webpack. I think this should resolve the need to know on what platform you're running. Can you give it a try? I would love to hear your feedback.

josdejong:
Alex, I've just published `workerpool@5.0.0`, which has a new option `workerType` (replacing the old `nodeWorker`). This allows you to explicitly select the type of worker you want without any magic, giving you full control. I think this makes integrating an environment variable redundant (if you want you could read the option from an environment variable that you create yourself though, so you have all freedom). What do you think?

Issue #36, opened 2018-01-10
Child processes not killed when the parent process is killed


dlecocq:
This is an issue that is relevant to `workerpool`, but it's not clear whether it should be the responsibility of the module or not.

When the parent process dies ungracefully, none of the spawned child processes are killed. For my particular use-case, I want to keep around a pool of long-running processes to perform work. If the parent process is killed with, say, `SIGTERM`, the child processes continue to run. When using a process management system like `upstart`, it's entirely possible that the parent process could be respawned, creating a whole new generation of worker processes in addition to the orphaned processes.

Here is a script that demonstrates the issue:

```javascript
#! /usr/bin/env node

const workerpool = require('workerpool');
const count = 4;
const pool = workerpool.pool({
  minWorkers: count,
  maxWorkers: count,
});

// This is meant to represent a long-lived process
const work = (id) => {
  const workForever = (id) => {
    console.log(`Worker ${id} alive and kicking...`);

    return new Promise((resolve) => {
      setTimeout(resolve, 1000);
    }).then(() => workForever(id));
  };

  return workForever(id);
};

const ids = Array.from(Array(count).keys());
const promises = ids.map((id) => pool.exec(work, [id]))

Promise.all(promises)
  .then(() => console.log('Done with all workers'));

// After dispatching all the child processes, simulate the parent process getting
// killed by a user, out-of-memory error, etc.
process.kill(process.pid, 'SIGKILL');
```

While most signals could be caught with `process.on` to then explicitly terminate the pool, `SIGKILL` does not provide such an opportunity. After some googling, I found a solution that may be of help to others who encounter similar situations. Apparently the `'disconnect'` event available in `process` is triggered when the parent process dies, thus making it possible to have the child processes self-terminate. The updated `work` function then becomes:

```javascript
const work = (id) => {
  process.on('disconnect', () => {
    process.exit(0);
  });

  const workForever = (id) => {
    console.log(`Worker ${id} alive and kicking...`);

    return new Promise((resolve) => {
      setTimeout(resolve, 1000);
    }).then(() => workForever(id));
  };

  return workForever(id);
};
```

josdejong:
Thanks for bringing this up Dan, and for your clear demonstration. I think the ideal situation is when workerpool neatly kills it's children when being killed itself, there is no one else who can take ownership of the orphaned childs.

Would it be possible for you to work out your solution in a pull request? That would be very helpful!

dlecocq:
[...]

PR #80, opened 2019-07-25
Worker exit when pool(parent process) is killed

RogerKang:
fix this issue: Child processes not killed when the parent process is killed #36 
## Implementation
Worker listen to 'disconnect' event to exit, when parent process is killed, worker will exit subsequently.
## Test
Fork a child prcocess to use workerpool (which will invoke an interval child worker), then kill this child process to detect whether the worker(child process of the child process) is exited in a short time.


josdejong:
Thanks a lot for this fix @RogerKang!

PR #84, opened 2019-08-20
Missing nodeWorker option in _ensureMinWorkers method

spacelan:
If the minWorkers option is set, workerpoll will always creates child processes at initialization, even though the nodeWorker option is `thread` or `auto`.

Ref: https://github.com/josdejong/workerpool/commit/571357e3e08007a5cec5442cb53b0b65e077f4e9#diff-1b589a05fcf1509343500faf9cd70bdaR247


josdejong:
Thanks for this fix @spacelan !

It may be even better to pull out the two cases of `new WorkerHandler(...)` to prevent future mismatches.

PR #91, opened 2019-09-14
Fix README

tamuratak:
`'auto'` is default for `workerType`.


josdejong:
Thanks for the fix Takashi!

PR #100, opened 2019-10-19
Fix typo

dandv:
\+ pinging the folks over at https://github.com/developit/workerize/issues/37 since that serialization implementation seems to be broken


josdejong:
Thanks for fixing this typo Dan :)

PR #106, opened 2019-11-02
Fix environment detection in browser

acgrid:
`typeof process.versions != null` will fail if `process` is a defined variable, for example, browser environment with improper webpack bundled

### DIFF
--- a/src/environment.js
+++ b/src/environment.js
@@ -1,13 +1,14 @@
 var requireFoolWebpack = require('./requireFoolWebpack');
 
 // source: https://github.com/flexdinesh/browser-or-node
-module.exports.isNode = isNode = function (nodeProcess) {
+var isNode = function (nodeProcess) {
     return (
         typeof nodeProcess !== 'undefined' &&
         nodeProcess.versions != null &&
         nodeProcess.versions.node != null
     );
 }
+module.exports.isNode = isNode
 
 // determines the JavaScript platform: browser or node
 module.exports.platform = isNode(process)
--- a/src/environment.js
+++ b/src/environment.js
@@ -1,13 +1,16 @@
 var requireFoolWebpack = require('./requireFoolWebpack');
 
 // source: https://github.com/flexdinesh/browser-or-node
-var isNode = (
-    typeof process !== 'undefined' &&
-    process.versions != null &&
-    process.versions.node != null);
+module.exports.isNode = isNode = function (nodeProcess) {
+    return (
+        typeof nodeProcess !== 'undefined' &&
+        nodeProcess.versions != null &&
+        nodeProcess.versions.node != null
+    );
+}
 
 // determines the JavaScript platform: browser or node
-module.exports.platform = isNode
+module.exports.platform = isNode(process)
     ? 'node'
     : 'browser';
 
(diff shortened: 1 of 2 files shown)


acgrid:
I am not clear about the cause of failed test in Node 8.
Seems that mocking the `global` would break things in Node 8, any help is welcome.

josdejong:
Thanks for this fix @acgrid! I will have a look to see what's going wrong with node 8 (let's not bother too much though, node 8 has end of life in less than 2 months).

josdejong:
@acgrid I think the issues with node 8 arise from the tricks with `requireUncached` and replacing global things. 

Maybe instead of trying to get the environment simulated correctly, we can make the code better testable. If we change `isNode` into a function with `process` as argument, it will be trivial to test. Something like:

```js
function isNode (nodeProcess) {
  return (
    typeof nodeProcess !== 'undefined' &&
    nodeProcess.versions != null &&
    nodeProcess.versions.node != null
  );
}

module.exports.platform = isNode(process)
    ? 'node'
    : 'browser';

module.exports.isNode = isNode;
// ... now it's easy to test isNode(...)
```
What do you think?

acgrid:
You're right!
We can just making the function as a pure function. So it is much better to be tested.

josdejong:
Ok cool. Do you want to give this a try and change `isNode` into a function? I can do it too if you want.

acgrid:
Please check again, thanks!

josdejong:
Looks good, thanks for the updated PR.

Two small feedbacks:

1.  Can you please define `isNode` explicitly as a variable (const or let), it currently is implicitly defined. Or alternatively, simply call `module.exports.isNode` instead of this local variable.
2.  I think the unit tests can be simplified. The constants `test`, `shouldBeBrowser`, `shouldBeNode` introduce some indirections that make it hard to see what the tests are really doing. Instead, we can simply call `isNode` in a test, like:

    ```js
    describe('should distinguish environment', function () {
    
      it('...', function () {
        assert.strictEqual(isNode(...), false)
        assert.strictEqual(isNode(...), true)
        // ...
      })
    
    })
    ```
    What do you think?

josdejong:
wow this is plain and simple :sunglasses: I like it.

Thanks again for the fix!

josdejong:
@acgrid your fix is now published in `v5.0.2`

acgrid:
Roger, thanks!

tteke:
I don't know if this was tested on the browser but as far as I can see it does not work. As process is not defined, it immediately throws instead of getting nodeProcess as undefined.
Also, isMainThread function is also broken it just checks if isNode function is defined ( it does not invoke that function ). I think this should be reverted because if webpack is bundled improperly, thats not the concern of this package. Also because it does not work in the browser at all.

josdejong:
Thanks for reporting @tptee apparently we didn't do proper testing on the browser :scream: 

I've immediately created a fix (ab74b1410a94a1cdfb2685638c1d81c037a263d0) and published as `v5.0.3`. 

The `isNode` function is only when detecting the environment automatically. This automatic detection may indeed not always work correctly depending on Webpack configuration. To overcome this you can simply specify a `workerType` for either browser or nodejs instead of rely on automatic detection.

tteke:
Thanks for the quick fix. I will assume that thanks was meant for me and take it :)

PR #143, opened 2020-04-08
Add Extra functionality to the pool

JPig:
- Added extra option to pool constructor to allow disabling `autoStart`
- Added ability to `pause` workerpool
- Added ability to `resume` workerpool
- Added ability to `enqueue` tasks into the pool/queue seperate to worker
- Added tests for all added functionality

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,9 @@ The following options are available:
   - In case of `'web'`, a Web Worker will be used. Only available in a browser environment.
   - In case of `'process'`, `child_process` will be used. Only available in a node.js environment.
   - In case of `'thread'`, `worker_threads` will be used. If `worker_threads` are not available, an error is thrown. Only available in a node.js environment.
+- `autoStart: boolean`. Define whether the worker pool should start processing tasks by default, or if it should be paused initially.
+  - In case of `true` (default), the worker pool will start processing tasks as soon as they are added/given to the pool
+  - In case of `false`, the worker pool will default to paused, and can be resumed using `Pool.resume()`
 
 A worker pool contains the following functions:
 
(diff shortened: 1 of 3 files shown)


josdejong:
Thanks Jordan for your PR! 

Can you explain what `enqueue` does exactly? Isn't it possible to use the existing API `pool.exec` which can keep tasks in the queue when the pool is paused?

These new methods add bloat to the API of pool. That can be worth it, but it's good to think critically what the added value is and whether this is serving a generic or a very specific use case. A thought: Would it be possible to use composition here instead of extending `Pool`? It may be possible to create this pauseable queue as a wrapper around the workerpool.

JPig:
in regards to `enqueue` I guess it's just in my eyes a helper function in the case of needing to batch add multiple tasks to the workerpool, It just makes it a bit nicer than a loop of multiple function calls etc. Im not really in any way tied to this part of the functionality though.

And in regards to the added `pause`/`resume` and `autoStart` functionality that was more for a utility to make the library more testable, just the ability to have a bit more fine grained control over what happens when.

And in terms of the composition thought, do you mean the like refactor the PR and the wrapper would be inside of this package/library as say an alternate class that can be used?, or are you suggesting to just add said functionality within the code of where its being implemented etc?

josdejong:
Thanks for the clarification Jordan 👍 . Again, I'm trying to find a good balance between a simple, non-bloated API, and useful functionality.

I have the feeling that the `enqueue` method adds quite some code and relatively little added value. Unless I'm overlooking something I would think you can achieve the same with this helper function:

```js
function enqueue (pool, tasks) {
  tasks.forEach(task => pool.exec(task.method, task.params) 
}
``` 

I think there could be value in adding a _pause_ functionality, but I would like to have a stronger case before integrating it. I understand it can make testing easier, but without real world examples of needing this for real in an _application_ I would like to put this feature on hold. 

On a side note, what I'm a bit afraid of is that when implementing "pause", the expectation of users will be that tasks are _really_ paused. However, as you clearly explain in the docs, the pausing only prevents starting new tasks, but cannot pause currently running tasks. Depending on real-world use cases this may be fine or turn out to be a non-useful solution in practice.

What do you think?

josdejong:
@JPig what do you think? Shall we close this PR?

JPig:
> @JPig what do you think? Shall we close this PR?

Hey sorry for the slow reply, yep that is fine with me.

josdejong:
:+1: thanks for your input and efforts!

PR #151, opened 2020-05-12
export source for node and unminified for browser

boneskull:
Hi,

over at [Mocha](/mochajs/mocha) we're looking to add support for parallel testing (mochajs/mocha#4245), and I've chosen `workerpool` to help manage this for us.

I noticed when debugging that this project is shipping a `.min.js` file to Node.js users.  This isn't necessary, and can make debugging awkward.

This PR changes the main export in Node.js to be the source entry point: `src/index.js`.  I've added the `browser` field to point to `dist/workerpool.min.js`.

Also:

Those building browser apps will often do their own minification.  For those apps which do not, it can make sense to ship a minified version--but it doen't need to be the main export.  I _have not changed_ the default browser export to `dist/workerpool.js` (which it likely should be) because it too needs a source map (pointing back to the files in `src/`), and I wasn't sure how to configure that in the gulpfile.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,8 @@
     "pool",
     "isomorphic"
   ],
-  "main": "dist/workerpool.min.js",
+  "main": "src/index.js",
+  "browser": "dist/workerpool.js",
   "files": [
     "dist",
     "examples",


boneskull:
This _could_ be a breaking change for some consumers who bundle `workerpool` and are not looking in the `browser` field for the entry point.

josdejong:
Thanks for your PR Christopher, that sounds like an awesome feature. I use mocha in quite some projects and parallel testing would be great 👍 . There is almost a circular dependency then since `workerpool` also uses `mocha` 😄 , it's only a `devDependency` though so no problem.

I didn't know about not having to provide minified code, there is a bit too much going on in js bundling world to follow everything. Still have to catch up with ES6 modules and everything. I will take some time tomorrow to merge your PR and change the gulpfile to also generate a map file for the non-minified bundle, and then it's fine to point to the non-minified `dist/workerpool.js` from the `browser` field. 

Will be a breaking change indeed so I can publish it as `v6.0.0`, hopefully tomorrow after testing from an external browser and node.js project.

josdejong:
@boneskull I've published `workerpool@6.0.0` containing the updated entry points in `package.json`.

PR #158, opened 2020-06-06
Include the pools process ID (poolID?) in the pools stats

trpapp:
Stats are a great way to show what is happening within a pool but it should also contain a reference (process ID) to the pool itself.

### DIFF
--- a/src/Pool.js
+++ b/src/Pool.js
@@ -315,15 +315,17 @@ Pool.prototype.terminate = function (force, timeout) {
 
 /**
  * Retrieve statistics on tasks and workers.
- * @return {{totalWorkers: number, busyWorkers: number, idleWorkers: number, pendingTasks: number, activeTasks: number}} Returns an object with statistics
+ * @return {{processID: number, totalWorkers: number, busyWorkers: number, idleWorkers: number, pendingTasks: number, activeTasks: number}} Returns an object with statistics
  */
 Pool.prototype.stats = function () {
+  var processID = process.pid;
   var totalWorkers = this.workers.length;
   var busyWorkers = this.workers.filter(function (worker) {
     return worker.busy();
   }).length;
 
   return {
+    processID:     processID,
     totalWorkers:  totalWorkers,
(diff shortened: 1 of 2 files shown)


trpapp:
Oh, this may be great for NodeJS but isn't compatible with browsers.

trpapp:
If this is something the community is interested in, I'd we willing to implement a similar paradigm that includes this functionality within browsers as well. My thought process is that a "poolID" may be useful for multiple workerpools such that if it's within a NodeJS environment, the poolID is equivalent to the processID found by using process.pid; while in the browser, it would be referenced via an incremental counter (every time a new Pool is created).

Some questions on my mind:
-Is this scope creep?
-Is this best served by other implementations?

Thanks,
Tom

josdejong:
Thanks Thomas for your PR. 

I'm not entirely sure what problem this PR solves, can you explain your use case? I would think that if you execute `myPool.stats()`, you already know which pool you're dealing with, namely `myPool`.

trpapp:
I guess I'm trying to solve a general problem, but it's more of a feature. My use case is to dynamically create these pools and have a way to reference/recognize them via a PoolID. More specifically, I'm creating a NodeJS cluster (based on how many logical cores I have) and each node in the cluster would have its own workerpool (or multiple pools). I'd like each pool to be able to report what process ID it "lives in". That is my current goal. Let me know what you think!

-Tom

josdejong:
I think you can quite easily write a wrapper around `workerpool` which creates an id for the pool and reports this along with statistics. I would like to keep `workerpool` as concise as possible. Something like:
```js
function createPoolWithId () {
  const poolId = process.pid
  const pool = workerpool.pool()
  return {
    pool,
    statsWithId: function () {
      const stats = pool.stats()
      return { ...stats, poolId }
    }
  }
}

trpapp:
Sounds good! I can write a wrapper, that's fair. Thanks for taking the time to review my branch! I'll close it off now.

josdejong:
:+1:

PR #159, opened 2020-06-10
Add promises-aplus tests for the promise library

forivall:
I was curious if the promise implementation you have is Promises/A+ conformant, and unfortunately, it's not, since it appears to be more of a jquery promise-compatible, which does some things synchronously that should be async.

The results summary:
>  266 passing (2m)
>  592 failing

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2324,6 +2324,15 @@
         "for-in": "^1.0.1"
       }
     },
+    "formatio": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/formatio/-/formatio-1.1.1.tgz",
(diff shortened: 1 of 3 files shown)


josdejong:
Thanks for your inputs Emily. The reason we couldn't use a standard Promise library (or nowadays the Promise implementation built in JS itself) is that we need support for cancellation. So I think it will not be possible to get the workerpool version of Promise standards compliant, though we could set some steps in that direction I guess. 

What we could do is rename the (internal) `Promise` class to `CancellablePromise` or something like that to make it more clear that it's not a standard Promise implementation.

What do you think?

josdejong:
It would be nice to see if we can replace Promise implementation with an external library like `p-cancelable`. I'm not sure if there would be incompatibilities and if it's possible to achieve or not, but it may be worth an experiment. Anyone interested in trying this out?

Issue #147, opened 2020-04-25
TypeError: Cannot set property 'ready' of null


velut:
Hi,

Sometimes when testing a [NestJS service class](https://docs.nestjs.com/providers#services) holding a pool to a dedicated worker, tests fail with the error below.
Most of the time, however, the same tests pass correctly.

```
/myProject/node_modules/workerpool/dist/workerpool.min.js:25
[...]

PR #160, opened 2020-06-11
Disregard messages from terminated workers.

Madgvox:
Fixes https://github.com/josdejong/workerpool/issues/147.

Will add a test at some point.

### DIFF
--- a/src/WorkerHandler.js
+++ b/src/WorkerHandler.js
@@ -203,6 +203,8 @@ function WorkerHandler(script, _options) {
   // queue for requests that are received before the worker is ready
   this.requestQueue = [];
   this.worker.on('message', function (response) {
+    if( me.terminated ) return;
+
     if (typeof response === 'string' && response === 'ready') {
       me.worker.ready = true;
       dispatchQueuedRequests();


josdejong:
This looks like a proper fix Nate, thanks!

Would be great if you can write a unit test for it so we prevent any future regressions.

josdejong:
@Madgvox would it be possible to write a unit test? Would be great to get this merged.

Madgvox:
Coming back around to this -- I will try to create a failing test case for it. This is a tricky one since the circumstances that cause the event to be fired after the worker has been terminated are not clear.

josdejong:
Thanks!

hhprogram:
I see this PR is still open. Is there a particular reason this hasn't been merged in yet? Is it that the relevant unit tests haven't been written yet? I can try to make some good ones, but I agree with @Madgvox  - I'll have to think about what a good unit test for this would be, since I've not been able to reproduce this consistently. 

running into similar issues as #147 with unit tests using jest.

josdejong:
Thanks @hhprogram for your input! 

It would indeed be nice to have a unit test, if this is not _too_ complicated of course. Besides that I think the PR itself is just fine (one of the smallest PR's ever 😉).

Do you think it's possible to write a test for this?

hhprogram:
> Thanks @hhprogram for your input!
> 
> It would indeed be nice to have a unit test, if this is not _too_ complicated of course. Besides that I think the PR itself is just fine (one of the smallest PR's ever 😉).
> 
> Do you think it's possible to write a test for this?

@josdejong  Sorry just saw this response - I'm working on it. Will try to in the next day or so (or comment with an update of the progress I was able to make)

Was thinking about it more though, does it make sense to do the check implemented here, or does it make sense to add logic to the `terminate` function and then close the event listeners in that function? This is an open question - as I'm no JS expert

Madgvox:
Thanks for taking a look at this @hhprogram. Life gets in the way as I'm sure you know, and when I was poking around at this I could never figure out a way to reliably recreate the conditions under which it might happen organically, so it fell to the bottom of my todo list.

josdejong:
Thanks @Madgvox for your efforts! Will close this PR now in favor of #264 and #265.

Issue #175, opened 2020-08-04
workerpool does not wait for child processes to terminate


boneskull:
Hi,

It looks like when `Pool#terminate()` is called when using the `child_process` builtin, `ChildProcess#kill()` is used.  But this does not mean the child process _has actually exited_.  The current behavior may be fine in most situations--fire-and-forget--but I need to ensure the child processes have ended before moving forward.  Waiting _will_ have performance impact, so if this were to change, perhaps it's best as an option?

Anyhow, changing this relevant code in `WorkerHandler#terminate` seems to work, assuming `kill()` works as intended.  

```js
      if (typeof this.worker.kill === 'function') {
        this.worker.kill();  // child process
        this.worker.on('exit', () => {
          this.worker = null;
          this.terminating = false;
          this.terminated = true;
          if (this.terminationHandler) {
            this.terminationHandler(this);
          }
        })
        return; 
      }
```

I note that the return value of `this.worker.kill()` above is ignored; if the signal fails to send, `ChildProcess#kill` returns `false`--but there's no error-handling of that situation in `workerpool`.  Couldn't tell you what _should_ happen in that case, but just pointing it out. :smile:

This seems to differ from Node.js' `Worker#terminate()`, which looks like it's guaranteed to end the worker thread.

josdejong:
Thanks for reporting, there are indeed options to improve the `terminate` method.

I think this is a duplicate of #32. Help improving this would be very welcome, I think we should give the `terminate` method as a whole some love.

boneskull:
closing in lieu of #32

josdejong:
This should be fixed now in `v6.0.2` thanks to #176

PR #176, opened 2020-08-11
Pool#terminate() waits until subprocesses are dead to fulfill

boneskull:
- `Pool#_removeWorker()` is now async
- ensure task resolver handles cancellations & timeouts in sequence
- mock behavior of `subprocess.kill()` for worker threads (returns a boolean)
- mock `subprocess.killed` for worker threads (a boolean)
- many changes to tests to support async nature of terminate
- some tests changed to use native Promise-handling functionality of Mocha (where it made sense)

* * *

I spent quite a bit of time with this one, as I was struggling a bit with the builtin `Promise` implementation.  _Hopefully_ there are no unintended consequences here :smile:

Ref: #32
Ref: #175

### DIFF
--- a/src/Pool.js
+++ b/src/Pool.js
@@ -207,10 +207,6 @@ Pool.prototype._next = function () {
           .catch(function () {
             // if the worker crashed and terminated, remove it from the pool
             if (worker.terminated) {
-              // _removeWorker will call this, but we need it to be removed synchronously
-              me._removeWorkerFromList(worker);
-              // If minWorkers set, spin up new workers to replace the crashed ones
-              me._ensureMinWorkers();
               return me._removeWorker(worker);
             }
           }).then(function() {
@@ -259,17 +255,22 @@ Pool.prototype._getWorker = function() {
 };
 
 /**
- * Remove a worker from the pool. For example after a worker terminated for
- * whatever reason
+ * Remove a worker from the pool. 
+ * Attempts to terminate worker if not already terminated, and ensures the minimum
+ * pool size is met.
  * @param {WorkerHandler} worker
+ * @return {Promise<WorkerHandler>}
  * @protected
  */
 Pool.prototype._removeWorker = function(worker) {
-  DEBUG_PORT_ALLOCATOR.releasePort(worker.debugPort)
+  DEBUG_PORT_ALLOCATOR.releasePort(worker.debugPort);
+  // _removeWorker will call this, but we need it to be removed synchronously
+  this._removeWorkerFromList(worker);
+  // If minWorkers set, spin up new workers to replace the crashed ones
+  this._ensureMinWorkers();
   // terminate the worker (if not already terminated)
   return new Promise(function(resolve, reject) {
     worker.terminate(false, function(err) {
-      me._removeWorkerFromList(worker);
       if (err) {
         reject(err);
       } else {
--- a/src/Pool.js
+++ b/src/Pool.js
@@ -207,10 +207,13 @@ Pool.prototype._next = function () {
           .catch(function () {
             // if the worker crashed and terminated, remove it from the pool
             if (worker.terminated) {
-              me._removeWorker(worker);
+              // _removeWorker will call this, but we need it to be removed synchronously
+              me._removeWorkerFromList(worker);
               // If minWorkers set, spin up new workers to replace the crashed ones
               me._ensureMinWorkers();
+              return me._removeWorker(worker);
             }
+          }).then(function() {
             me._next(); // trigger next task in the queue
           });
 
@@ -264,8 +267,16 @@ Pool.prototype._getWorker = function() {
 Pool.prototype._removeWorker = function(worker) {
   DEBUG_PORT_ALLOCATOR.releasePort(worker.debugPort)
   // terminate the worker (if not already terminated)
-  worker.terminate();
-  this._removeWorkerFromList(worker);
+  return new Promise(function(resolve, reject) {
+    worker.terminate(false, function(err) {
+      me._removeWorkerFromList(worker);
+      if (err) {
+        reject(err);
+      } else {
+        resolve(worker);
+      }
+    });
+  });
 };
 
 /**
--- a/src/WorkerHandler.js
+++ b/src/WorkerHandler.js
@@ -391,12 +391,13 @@ WorkerHandler.prototype.terminate = function (force, callback) {
     if (this.worker) {
       if (typeof this.worker.kill === 'function') {
         // child process
-        // cleanup once the child process has exited
-        this.worker.once('exit', function() {
-          cleanup();
-        });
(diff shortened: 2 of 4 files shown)


boneskull:
Of note, I had to remove a few assertions which no longer made sense.  `Pool#terminate()` will kill the subprocesses _then_ fulfill, but if a promise is _canceled_ (or it times out, or the task fails, etc.), the subprocesses are _not_ killed before rejection of the `Promise` returned by `exec`.  

In other words, when `catch`-ing a canceled `Promise`, you are not guaranteed _n_ workers within the `Pool` instance.  

I don't particularly _like_ how that works, but changing it would require a somewhat gnarly refactor, and I'm not sure it's even desired behavior.  

For that reason, this PR is probably a breaking change.

josdejong:
Thanks a lot Christopher, this looks good on first sight! I'll review your PR asap, I hope coming weekend.

boneskull:
I think what's worrisome is that the tests appear to mainly target worker threads instead of child processes, unless I missing something... would seem that some portion of the tests should be repeated across all supported worker implementations

src/Pool.js
```diff
@@ -264,8 +267,16 @@ Pool.prototype._getWorker = function() {
 Pool.prototype._removeWorker = function(worker) {
   DEBUG_PORT_ALLOCATOR.releasePort(worker.debugPort)
   // terminate the worker (if not already terminated)
-  worker.terminate();
-  this._removeWorkerFromList(worker);
+  return new Promise(function(resolve, reject) {
+    worker.terminate(false, function(err) {
+      me._removeWorkerFromList(worker);
```
josdejong, line 272: With the new solution, termination is async, so there is a temporary situation where the worker is still in the worker list but is being terminated. Should we change this to remove the worker from the list first, and after that call .terminate?

src/Pool.js
```diff
@@ -207,10 +207,13 @@ Pool.prototype._next = function () {
           .catch(function () {
             // if the worker crashed and terminated, remove it from the pool
             if (worker.terminated) {
-              me._removeWorker(worker);
+              // _removeWorker will call this, but we need it to be removed synchronously
+              me._removeWorkerFromList(worker);
               // If minWorkers set, spin up new workers to replace the crashed ones
               me._ensureMinWorkers();
+              return me._removeWorker(worker);
```
josdejong, line 214: Maybe we can change this to something like:

```js
return me._removeWorker(worker).then(function () {
  // If minWorkers set, spin up new workers to replace the crashed ones
  me._ensureMinWorkers();
})
```

So we don't have to clear the worker from the list first? Having to do that feels a bit off, should ideally be only needed to call from within `_removeWorker`.

src/WorkerHandler.js
```diff
@@ -377,23 +377,38 @@ WorkerHandler.prototype.terminate = function (force, callback) {
...
         throw new Error('Failed to terminate worker');
       }
-      this.worker = null;
-    }
-    this.terminating = false;
-    this.terminated = true;
-    if (this.terminationHandler) {
-      this.terminationHandler(this);
     }
+    cleanup();
```
josdejong, line 411: It looks like `cleanup()` is called twice: one time right away, and one time on callback of `this.worker.once('exit'`. Should we split this in two functions? Maybe set `this.terminating = true` when starting to terminate, and `this.terminated = true` after termination finished?

josdejong:
Thanks a lot Christoper for all your work! It looks like a solid solution. 

I've placed a couple of feedbacks inline in the code.

The custom Promise implementation is not ideal, sorry about that. It's needed in order to be able to cancel the promise after a timeout or custom action. What would be nice is if we can make the behavior more standard compliant to have less WTF's when working with it. Or maybe we can come up with a totally different solution which doesn't require extending Promise with cancellation. I'm open to ideas in this regard!

About the unit tests for child_process vs worker_threads: that's a good point. Both should be tested thoroughly of course. It's currently a bit implicit. Maybe we can rewrite the relevant unit tests such that they are executed twice: once for 'process', and once for 'thread'.

josdejong:
@boneskull I see this PR is still open. Do you expect to have time to finish it?

boneskull:
Yes

src/Pool.js
boneskull: Are you asking me?  :smile:  I was keeping the series of instructions the same; terminate _then_ remove.  

However, the only place in the codebase that calls `_removeWorker()` removes it synchronously beforehand:

https://github.com/josdejong/workerpool/blob/57e6795ca08d51022847f8de204fca7ae0fb8407/src/Pool.js#L211-L214

It appears the call to `_removeWorkerFromList()` here is not needed unless you expect `_removeWorker()` to be called for some other reason.

src/Pool.js
boneskull: OK, I should probably read all the comments before replying.  But yes.

boneskull: Actually... no.  I think that would have an additional negative affect on performance--you have to wait for termination before spinning up a new one.  I think that just leaving this as-is and removing the call to `_removeWorkerFromList` in `_removeWorker` makes more sense.

boneskull: Or just moving it to before the `Promise`.  Let's just do that...

src/WorkerHandler.js
boneskull: It's not called twice unless somehow the child process has already been killed, then emits an additional `exit` event.  I have no idea if this is possible:

https://github.com/josdejong/workerpool/blob/57e6795ca08d51022847f8de204fca7ae0fb8407/src/WorkerHandler.js#L392-L401

However, that should probably change to:

```js
      if (typeof this.worker.kill === 'function') {
        // child process
        if (!this.worker.killed && !this.worker.kill()) {
          cleanup(new Error('Failed to send SIGTERM to worker'));
        } else {          
          // cleanup once the child process has exited
          this.worker.once('exit', function() {
            cleanup();
          });
        }
        return;
```

boneskull:
@josdejong I've sent an update, please take a look.

src/WorkerHandler.js
boneskull: FWIW I'm not entirely confident that just listening for the `exit` event is enough.  Maybe a strategy like [signal-exit](https://npm.im/signal-exit) would be more robust--_if_ that's possible with a child process.

src/Pool.js
josdejong: This makes sense 👍

src/WorkerHandler.js
josdejong: 👍 I think you're right

josdejong:
Thanks for the updates Christopher, looks really solid now, will merge your improvements.

josdejong:
Your fixes are published now in `v6.0.2` 🎉

Issue #51, opened 2019-02-22
Send progress/status information from worker to main process


dimitriylol:
As I understand, there isn't a way for listening/sending messages between worker and executor. I refer to message listeners and sender methods `worker.send` in `worker.js` and `WorkerHandler.js` files. May you give a suggestion how I can establish message communication? Maybe you've already had some thought about it.

P.S. Great package, thank you for the work.

josdejong:
Thanks. I indeed haven't thought about opening up two way communication between the main process and the workers. `workerpool` creates a "remote procedure call" on top of the bare-bone message passing mechanism that is the basis of the underlying workers.

We could discuss how allowing 2way message passing could look like, though I'm not sure if this fits in the scope of `workerpool` (sometimes it's best to have a library do just one thing). Do you have any ideas?

dimitriylol:
I give some thoughts about my specific task and realized that I want to implement [server sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) with message passing mechanism of the underlying workers instead of HTTP. So master process is client and worker is server which may produce messages to a master.

It can be done in two way: 
1. Quick and easy. Create new API function (let's call it `execWithMsg` which calls internally `exec`) and pass a callback for listening messages from worker. This callback is passed down to a callback in WorkerHandler.js file.
```javascript
this.worker.on('message', function (response) {
// call it here via task.msgCallback
}
```
And pass carried function for sending message from worker code:
```javascript
method.apply(method, [ ...request.params, (msg) => worker.send({ id: request.id, msg }) ]);
```
I've done rough implementation, so it's easy. This approach violates logic of "one function do just one thing" which is pretty bad. But if it's okay to you, I can create pull request in couple days.

2. Create an API function which implements server sent events pattern in addition to RPC. There can be a function `establishConnection`:
```javascript
function establishConnection (method, callbackForWorkerEvents) {
// in the same manner as `exec` creates a worker
// but returns undefined. All logic for processing messages from worker must be implemented in `callbackForWorkerEvents`
}
```

By the way, if someone would like to have 2way connection, then `establishConnection` must return `send` function instead of undefined. The function would send messages from master to client . Channel for listening messages from master can be done by writing following code in dedicated worker script:
```javascript
process.on('message', (msg) => {})
```
It should be abstracted in some way, but I didn't come up with any thoughts except passing stringified message listener to a worker script. IMO, it's not acceptable.

[...]

PR #210, opened 2020-11-26
Enable two way communication with worker using Event Emitter pattern

S0c5:
## Description

This PR enables 2-way communication between the main process with the task executed in the workers using an event emitter pattern, now we can listen and emit events 🔥 

```javascript


function processImage(url) { // some code that process a image asyn
  return new Promise(async (resolve) => {
    this.emit('progress', { completion: 0.5 }); // the Function was overloaded with emit, on, once
    this.on('pleaseFinishTask', function() { // this Example waits for this event to complete the task
      resolve(url);
    });
    await delay(5000);
    this.emit('progress', { completion: 1 });
 });
}

const poolInstance = pool.exec(processImage, ['https://someimage.com/url'])

poolInstance.on('progress', ({  completion }) => {
  const percentage =  completion * 100;
  console.log('Image processing Completion', percentage)
  if (percentage == 1) {
    poolInstance.emit('pleaseFinishTask')
  }
});

poolInstance.then((r) => {
  console.log('Result', r);
});
```


josdejong:
Thanks David! 

So this PR implements #51, right?

I will review your PR soon, all looks good and well tested at first sight. 

I see the tests fail on Node 10 with the following error though, can you have a look into that?

```
1) Pool
       should handle crashed workers (1):
     Uncaught TypeError: Cannot set property 'killed' of null
      at ChildProcess.<anonymous> (src/WorkerHandler.js:434:30)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:248:12)
```

See: https://travis-ci.org/github/josdejong/workerpool/jobs/746149617

S0c5:
Hello Jos!

S0c5:
Hello Jos,

Yeah, this PR implements and closes the #51, I didn't have time to see what happens on node 10 but maybe I can check this weekend. 

Let me know if you have comments on my code.

Regards.

josdejong:
David I've just had a look at the API you've implemented for passing messages. Overall it looks good but I think we should align a bit more on the API.


1.  API

    The worker side now has methods `on`, `emit`, `once` attached to `this` of the function. This feels tricky to me. I would prefer a more explicit API, like the last part in this proposal of @boneskull, see https://github.com/josdejong/workerpool/issues/51#issuecomment-637087443:

    ```js
    // worker.js

    // worker is an EventEmitter
    let worker = workerpool.worker({
      myFunc: async () => {

        // emit() could attach some metadata about the worker to its payload; perhaps unique ID / pid
        worker.emit('<user-defined event name>'), data);
      }
    });
    ```

    About the main process side: I think we should bind the events to a specific task and not to the pool, like I explain in https://github.com/josdejong/workerpool/issues/51#issuecomment-641793438: 

    ```js
    // main.js

    const result = await pool.exec('fibonacci', [10], {
      // pass an event listener along when starting a task
      on: function (event, data) {
        // event callback can be used for anything. 
        // For example event can be 'progress', and data can be a percentage like 0.25
        // maybe we should also pass the original method and arguments so you have all
        // context available
        console.log(event, data)
      }
    })
    ```

    What do you think?

2. It is an interesting idea to also allow the main process to post messages towards the worker (with `this.on('pleaseFinishTask'` in your example). I guess this could be useful, but do you have concrete use cases for this? If we do not really need it, it would make the solution simpler and smaller.

3. Maybe instead of creating an eventEmitter ourselves, we could use a small, battle-tested event emitter from npm?

4. I do not fully understand what `__EVENT__` is doing and why it is needed. Can you explain?

PR #212, opened 2020-11-30
use `new Function` instead of eval

tjenkinson:
This is a bit safer and a bit more performance, because it means the function only has access to the global scope. Previously it would be able to access and rewrite things like `TERMINATE_METHOD_ID`.


josdejong:
Thanks Tom, good idea 👍 

The imports for the custom Promise is indeed redundant nowadays and can be removed.

josdejong:
I've finally published this improvement in `v6.0.4`. Sorry for the delay @tjenkinson .

PR #225, opened 2021-01-05
Allow passing of options to worker threads

akibnizam:
With this change, you can pass your own options for a Worker.

Just add all your custom worker thread specific options to the `workerThreadOpts` field of the `options` passed in Pools.

For example --

`const pool = workerpool.pool(__dirname + '/myWorker.js', {workerThreadOpts: {stdout: false, stderr: false}});`

If no workerThreadOpts passed in options of workerpool.pool, then the default `{stdout: false, stderr: false}` will be used.

Options that you can now pass to `workerThreadOpts` are -- https://nodejs.org/api/worker_threads.html#worker_threads_new_worker_filename_options

### DIFF
--- a/src/Pool.js
+++ b/src/Pool.js
@@ -29,7 +29,6 @@ function Pool(script, options) {
   this.nodeWorker = options.nodeWorker;
   this.workerType = options.workerType || options.nodeWorker || 'auto'
   this.maxQueueSize = options.maxQueueSize || Infinity;
-  this.workerThreadOpts = options.workerThreadOpts || {};
   this.processOrWorkerOpts = options.processOrWorkerOpts || {}
 
   // configuration
--- a/src/Pool.js
+++ b/src/Pool.js
@@ -29,6 +29,7 @@ function Pool(script, options) {
   this.nodeWorker = options.nodeWorker;
   this.workerType = options.workerType || options.nodeWorker || 'auto'
   this.maxQueueSize = options.maxQueueSize || Infinity;
+  this.workerThreadOpts = options.workerThreadOpts || {};
 
   // configuration
   if (options && 'maxWorkers' in options) {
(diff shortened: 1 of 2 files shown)


josdejong:
Thanks for your PR Andrew. Makes sense to be able to specify `stdout` and `stderr`. Just to get a better picture, can you explain your use case for this?

Your current PR only applies those options for `WorkerThreads`, not for `child_process`. Would it make sense to make those options more generic and also implement this for `child_process`?

Before the PR is ready, I think it should have some unit tests to verify the behavior, and have documentation of this new feature.

akibnizam:
Hi Jos, thanks for the suggestions. Also my name is Akib, not Andrew. 

Good point about making it available for `child_process` too, so I added the functionality to make it usable for `worker` or `process`. 

My use case for allowing custom options for `worker_threads` in the first place was to specify `resourceLimits` options of worker threads. 

Also, this new modification will work with `child_process` only if the user specifically sets `workerType` as `process`. Since the options for `child_process` and `worker_thread` are not the same.

josdejong:
> Also, this new modification will work with child_process only if the user specifically sets workerType as process. Since the options for child_process and worker_thread are not the same.

Hm that's a good point. So currently we would allow the user to pass an object `processOrWorkerOpts`, and this is only applied when `workerType` is `process` or `thread`. If child_process and WorkerThreads really need different options, it may be not a good idea after all to try unify this into one option `processOrWorkerOpts`. Would it make sense to introduce two explicit options instead, like `processOptions` and `threadOptions`?

Two more feedbacks:

1. Can you describe the new options in the README?
2. Is it possible to write a unit test for this new functionality?

akibnizam:
Hi Jos, I got caught up with some personal work. You made some great points and I will address them very soon!

I will write new options in the README for sure

josdejong:
Thanks for the update, take it easy 👍

josdejong:
Will close this PR in favor of #296. Thanks for the work you've put into it Akib.

PR #243, opened 2021-02-23
chore: fix package bundling

Nytelife26:


josdejong:
Thanks for your PR @Nytelife26 . 

Can you explain why you need the source code under `src` over the bundle in `dist`?

josdejong:
O wait, I did read your PR wrongly, you're suggesting to _remove_ the code in `src` from the bundle. That makes sense to me indeed. But let's publish that as a breaking change.

Nytelife26:
@josdejong Why? It isn't a breaking change. It removes nothing for the end-user, just takes the source code out of the bundle to trim size and exclusively delivers the distribution channel.

josdejong:
That is true: it will only remove something that is undocumented, so it _should_ not break anything for anyone. I've become cautious with such cases over the years though, I've learned that users are extremely creative and will use any undocumented features if needed to workaround issues they come across. So, to prevent nasty surprises for anyone, I prefer upping the major version number.

Nytelife26:
@josdejong The source being bundled is not an undocumented feature. It is not a feature at all. Anyone that uses the source code from the archive to do anything with is most likely not using that in their application at all, since node resolves to the distribution channel of the bundle. The only possible thing I can see is something where they should be using the repository to get the source code anyways, and so this doesn't need to be anything more than a patch change for package metadata.

But, ultimately I guess it isn't my project. I just hope you're aware of the drawbacks and unintended consequences waiting for a major version to do this may have.

josdejong:
> I just hope you're aware of the drawbacks and unintended consequences waiting for a major version

I'm not sure what you mean, can you explain?

Nytelife26:
@josdejong As with any package, when a new major version is made, some installations will be left behind. Due to how large dependency graphs in node projects normally are, it will take at the very least a few months usually for the update to become the most distributed version, if at all. In the meantime, people looking to reduce bundle sizes and installation times either have to suffer or dynamically hook into their dependency resolution. This is all a rather large pain, as opposed to just implementing it as a change that npm and other package managers will switch to without complaints, as it is not breaking to the actual package.

josdejong:
That's a good point indeed, people are more reluctant to upgrade to a major (breaking) version because that can give issues, so that will lead to more applications using an outdated version for a longer period. 

I agree with you that it is a bit far fetched to see this (removing the `src` folder from the npm package) as a breaking change. I guess I may have grown a bit too conservative in this regard. In the past I've had too often that minor/bugfix upgrades of libraries I was using for unclear reason did break something whilst they should be backward compatible in theory. I've also regularly published libraries myself with a bugfixes or new features, being convinced that this was fully backward compatible, but sometimes this _still_ turned out to be breaking for some users in subtle ways. This often is related to the docs not documenting every detail. There are always assumptions in play, and there is always implicit behavior, en there are simply cases where people rely on unofficial workarounds to just get stuff working. So I'm a bit cautious in this regard.

Anyway, interesting discussion, but no need to overthink this 😄 . I'm ok with publishing this as a bugfix release then.

josdejong:
Published now as `v6.1.1`

DouglasDev:
This change has broken this package, at least on my machine: https://github.com/snowpackjs/snowpack/tree/main/plugins/plugin-babel

rwjblue:
Hmm, this leaves the package in a fundamentally broken state. The `main` entry point still points at `src/index.js` (and `npm publish` sees that this is your entry point and properly bundles it), but then any requires from that file do not work.

https://github.com/josdejong/workerpool/blob/0fb81b22b09d78eb14ce0053d0f428d6ac90cae0/package.json#L19

rwjblue:
Reported in #253

josdejong:
Ouch. The main file in package.json still pointed to `./src/index.js`. I should have seen that. I've reverted this change in `v6.1.2`, back to the drawing board.

PR #264, opened 2021-03-27
disregard messages from terminated workers

hhprogram:
credit / source: https://github.com/josdejong/workerpool/pull/160 

didn't have access so decided to make my own PR (more than happy to work with the Madgvox to get the unit test on the original PR so credit is appropriately given, but couldn't seem to @ the user here, so I'll let @josdejong, you decide what you want to do).

Anyways - like the linked PR says, solves [#147 ](https://github.com/josdejong/workerpool/issues/147)

After lots of digging, this unit test seemed best to me, in terms of being simple and targeted. Happy to discuss if you think this doesn't cover the case. 

Note: you can test this by commenting out the below code block found in `WorkerHandler.js` and this new unit test should fail.

```
if (me.terminated) {
    return;
}
```

potentially related to this [PR](https://github.com/josdejong/workerpool/pull/265) - maybe a good idea to review that linked PR before deciding on whether or not to merge this one in.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "workerpool",
   "license": "Apache-2.0",
-  "version": "6.1.2",
+  "version": "6.1.3",
   "description": "Offload tasks to a pool of workers on node.js and in the browser",
   "homepage": "https://github.com/josdejong/workerpool",
   "author": "Jos de Jong <wjosdejong@gmail.com> (https://github.com/josdejong)",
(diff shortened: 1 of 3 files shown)


hhprogram:
@josdejong let me know if the unit tests are comprehensive enough, if not, let me know what you think might be missing and I can work to add

Madgvox:
Good thinking @hhprogram -- when I was approaching this task I was thinking about it from the perspective of trying to mimic the conditions under which this occurred originally. It is much simpler to skip the complicated bit and simulate said conditions! If @josdejong gives the ok, I will copy your test over to my PR.

hhprogram:
@Madgvox sounds good! (Ya, I too overcomplicated it for awhile, and was spinning my wheels on how to unit test this, but luckily stumbled upon doing it this way after some head scratching).

josdejong:
Very smart idea to trigger this `ready` message yourself 😃

josdejong:
Closing this PR now, this fix is addressed now in #265.

PR #291, opened 2021-05-17
feat: worker handler last task options

meetlost:
**feature request**

I observed the workerHandler.processing contains working tasks, but in some circumstances (race condition), before the task is assigned to the worker, the worker handler have no idea about which task will be processing.

By the way, when trying to terminate workers manually, with this lastTaskOptions, we can decide which workers should be keep/kill.


josdejong:
Thanks for your PR @meetlost. I don't understand yet why you need this, can you explain a bit more how you use this new `lastTaskOptions`?

meetlost:
> Thanks for your PR @meetlost. I don't understand yet why you need this, can you explain a bit more how you use this new `lastTaskOptions`?

The `lastTaskOptions` can be used for filtering workers.

Suppose we have 10 workers doing 3 kinds of jobs, If we want to kill one kind of job (contain 3 workers as an example), as I understand the code, there is no way to do it right now. Pool.terminate() can terminate all workers though.

The task option can contain an unique job reference, then we can choose which workers to terminate.

josdejong:
The core of what `workerpool` tries to do for you is abstract away workers and queues, so you can focus on just firing your task and waiting until the result is returned. There should be no need to interact with a specific _worker_, you should interact with a specific _task_ instead. Currently you can give a task a timeout, and you can cancel a task. 

> If we want to kill one kind of job (contain 3 workers as an example), there is no way to do it right now as I understand the code

Can you elaborate? What is a "job" as you describe exactly? Isn't the current `cancel()` on the returned promise what you're looking for?

meetlost:
> Can you elaborate? What is a "job" as you describe exactly? Isn't the current `cancel()` on the returned promise what you're looking for?

A job is a reference which related to application logic.

Take file downloading as an example, one file is referenced by one unique job id/reference. File downloading can be split into multiple tasks (downloading via range), and tasks are going to be executed by workers.

As you mentioned, `cancel()` can be used to terminate one task, but the problem is how to find the tasks the application wants to terminate. The `workerpool` have a pool of tasks and workers, which can be mapped to a list of jobs, suppose the application wants to terminate a subset of the jobs, we need to find those `tasks` and `workers` at first.

I tried worker.processing, but seems this property is changing with the state of the running task, once the task finishes executing, the `processing[id]` is deleted. Under heavy load, the `workerpool` is running, and an event happens to terminate some jobs from the `workerpool`, race conditions can happen for finding jobs. `tasks` always contain the task options, but the `workers` does not.

josdejong:
Ok clear, so with a job you mean something consisting of multiple tasks. `workerpool` has no logic in place to deal with groups of tasks, a "job". You could write a layer _on top_ of `workerpool` to process a "job". The job can keep track of each of the tasks. When cancelling the job, it should cancel all of the tasks, and when one of the tasks errors, it needs to handle that in a certain way. What do you think?

meetlost:
> Ok clear, so with a job you mean something consisting of multiple tasks. `workerpool` has no logic in place to deal with groups of tasks, a "job". You could write a layer _on top_ of `workerpool` to process a "job". The job can keep track of each of the tasks. When cancelling the job, it should cancel all of the tasks, and when one of the tasks errors, it needs to handle that in a certain way. What do you think?

Yes, this is exactly what I want.
Seems adding a layer on top of `workerpool` can help me keep track of all the tasks. Will try this.

josdejong:
Ok nice! Would love to hear if it works out

josdejong:
Will close this PR now

PR #296, opened 2021-05-22
Allow forkOpts to be computed dynamically

forty:
Hello, 

I added this to be able use cases where I needed to attributed some unique identifiers to all workers in the pool.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,14 @@ The following options are available:
   - In case of `'web'`, a Web Worker will be used. Only available in a browser environment.
   - In case of `'process'`, `child_process` will be used. Only available in a node.js environment.
   - In case of `'thread'`, `worker_threads` will be used. If `worker_threads` are not available, an error is thrown. Only available in a node.js environment.
+- `forkArgs: String[]`. For `process` worker type. An array passed as `args` to [child_process.fork](https://nodejs.org/api/child_process.html#child_processforkmodulepath-args-options)
+- `forkOpts: Object`. For `process` worker type. An object passed as `options` to [child_process.fork](https://nodejs.org/api/child_process.html#child_processforkmodulepath-args-options). See nodejs documentation for available options.
+- `onCreateWorker: Function`. A callback that is called whenever a worker is being created. It can be used to allocate resources for each worker for example. The callback is passed as argument an object with the following properties:
+  - `forkArgs: String[]`: the `forkArgs` option of this pool
+  - `forkOpts: Object`: the `forkOpts` option of this pool
+  - `script: string`: the `script` option of this pool
+  Optionally, this callback can return an object containing one or more of the above properties. The provided properties will be used to override the Pool properties for the worker being created. 
+- `onTerminateWorker: Function`. A callback that is called whenever a worker is being terminated. It can be used to release resources that might have been allocated for this specific worker. The callback is passed as argument an object as described for `onCreateWorker`, with each property sets with the value for the worker being terminated. 
 
 > Important note on `'workerType'`: when sending and receiving primitive data types (plain JSON) from and to a worker, the different worker types (`'web'`, `'process'`, `'thread'`) can be used interchangeably. However, when using more advanced data types like buffers, the API and returned results can vary. In these cases, it is best not to use the `'auto'` setting but have a fixed `'workerType'` and good unit testing in place.
 
--- a/src/Pool.js
+++ b/src/Pool.js
@@ -23,12 +23,15 @@ function Pool(script, options) {
 
   options = options || {};
 
-  this.forkArgs = options.forkArgs || [];
-  this.forkOpts = options.forkOpts || {};
+  this.forkArgs = Object.freeze(options.forkArgs || []);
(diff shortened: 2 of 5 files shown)


forty:
I see that my test fails on the CI, but it fine locally, any idea what the issue might be?

josdejong:
Thanks Quentin for your PR. I've run your PR locally, all unit tests pass there too. I'm not sure why they are failing on the CI.

Can you explain your use case? Why do you need to know an unique id of every worker?

Related PR: #225

forty:
My very specific use case would be to be able to assign unique database to mocha worker processes, but I assume there could be other use cases.

The person who originally open the case wanted to assign unique port (though in that case, there probably other ways) https://github.com/mochajs/mocha/issues/4456#issue-713152909

More details:
- my comment on the same case detailing my use case https://github.com/mochajs/mocha/issues/4456#issuecomment-844924184
- a draft PR for the same user that opened the issue to solve the problem in mocha only https://github.com/mochajs/mocha/pull/4463/files (which is pretty clever if that indeed works, but is also an awful hack :grimacing: )

forty:
@josdejong what do you think of the use case? would you merge that PR? If yes, I can take some time to investigate the CI issue.

josdejong:
Sorry for the delay @forty. 

This use case makes sense indeed, so good to build a solution for it, thanks 👍 

Thinking aloud here: instead of passing dynamic options to a worker, wouldn't it make more sense to implement life cycle hooks for workers? So inside the worker you get two two methods `create` and `destroy`, and you can do initialization and cleanup work when needed, and from the workerpool you can keep track on created/destroyed workers, pass dynamic options via the create function. So in your case case when having a `destroy` function you could re-claim a "unique database" again after a worker did shutdown or crash.

forty:
@josdejong hi, I'm back :) I tried to put together something along the lines of you suggestions. It allows to release resources when the worker is terminated. Let me know if you like it, I'm happy to continue improving the PR if it's going in the right direction.

Thanks!

forty:
(in particular I know it's missing documentation and I'll update the typings)

josdejong:
@forty thanks for your updates. And sorry for my very late reply, I wasn't able to work much on my open source projects last months.

I really like where your PR is heading!! 😎 . I personally would rename `onWorkerCreation` and `onWorkerTermination` to `onCreateWorker` and `onTerminateWorker`, but if you like the current names better it is fine with me.

Indeed documentation and typings are missing, but besides that the PR looks ready to merge 👍

forty:
@josdejong no worries, thanks for getting back to me!

I didn't have strong feelings on the naming and like your proposal, so I made the update.

I added types and doc now, please let me know if that looks good. Thanks!

josdejong:
Wow that's fast 🚀 😄 , thanks, looks good indeed.

Before clicking the merge button, I would love hear if you have any thoughts on #329, which is related but about passing options to a worker in the browser. Do you have any ideas on how we could unify this @forty ? (or maybe we don't have to).

forty:
I think it would make sense to expose those new properties
1- in the pool options
and then
2- in the create/terminate Worker callbacks so they can be overriden as well for each worker

It should be an easy change once both PRs are merged.

josdejong:
That would be very flexible indeed. Ok let's first merge this PR and go from there. 

I see the CI tests are not running well for unclear reason (locally all tests work fine for me). I don't have time today but will look into it asap

forty:
I have rebased to latest master to get the GH workflow config, and I tested locally with this tool https://github.com/nektos/act, and I had no issue.
I'm not sure what to think :thinking:

forty:
@josdejong Found & fixed the issue :) 

(1) On GH, because the number of CPU is low, the default maxWorker is 1, which means that no parallelism happens.
Forcing it to a high enough number fixes the problem.

(2) The infinite loading is because the terminate function was never call in case a previous assert fails. Calling it in case of error fixes the problem.

Both problems could affect other tests. The second one is just annoying, but the first one could hide problems as many tests actually run with a single worker and no parallelism, it's possible that we could have tests failing locally and succeeding on GH.

Anyway, it should be all good now :+1:

josdejong:
Just wow. That makes sense, but what a tricky case! 

It would be nice if we could make sure the terminate function is always called, also in case of a failed assertion (like with `.finally(...)` in a promise). I'm not sure though if it is worth the trouble (I mean, normally the tests succeed and when one test fails and hangs, it's a good enough indication to have a look at that failed test).

Will merge your PR now, thanks again.

josdejong:
I added a usage example to the examples section: 5837a9cea3c320f28839b3f4de258c5be54b8c42, and have now published the new features in `v6.2.0` 🎉 .

forty:
Thanks!

Regarding the terminate pool in tests I also tried to think a bit how it could be done in a more consistent way.

What I think could work well would be to move the "var pool" one level higher and always re-use that same var. This way we could add a pool.terminate in a afterEach hook.

josdejong:
Ah, that makes a lot of sense, I think that will work for most (or all) tests!

PR #306, opened 2021-06-14
move to github actions

mansona:
I noticed when I checked test results on one of the PRs that Travis is saying `[they're] pushing back the shutdown date to June 15th`... That's very soon 😂 

I tried to keep this PR as close to what Travis was doing so that we could get it merged quickly and then iterate later. Feel free to take the code and push it to a branch on the main repo if it doesn't let you test the CI on this 👍

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,21 @@
+name: Node.js CI
+
+on: [push, pull_request]
+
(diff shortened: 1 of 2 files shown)


josdejong:
Thanks Chris! You're right, I still have to migrate a number of projects. Thanks for helping out here 👍

PR #230, opened 2021-01-29
Fix ERR_INVALID_ARG_TYPE on Node.js nightly

aduh95:
This PR fixes a bug that currently happen on nightly versions of Node.js due to a recent in the Node.js codebase:

```
1) WorkerHandler
       create a worker handler with custom script:
     Uncaught TypeError [ERR_INVALID_ARG_TYPE]: The "options" argument must be of type object. Received an instance of Array
      at ChildProcess.target.send (node:internal/child_process:716:7)
      at Array.forEach (<anonymous>)
      at dispatchQueuedRequests (src/WorkerHandler.js:262:21)
      at ChildProcess.<anonymous> (src/WorkerHandler.js:221:7)
      at ChildProcess.emit (node:events:378:20)
      at emit (node:internal/child_process:920:12)
      at processTicksAndRejections (node:internal/process/task_queues:82:21)

  2) WorkerHandler
       should handle the asynchronous initialization of a worker:
     Uncaught TypeError [ERR_INVALID_ARG_TYPE]: The "options" argument must be of type object. Received an instance of Array
      at ChildProcess.target.send (node:internal/child_process:716:7)
      at Array.forEach (<anonymous>)
      at dispatchQueuedRequests (src/WorkerHandler.js:262:21)
      at ChildProcess.<anonymous> (src/WorkerHandler.js:221:7)
      at ChildProcess.emit (node:events:378:20)
      at emit (node:internal/child_process:920:12)
      at processTicksAndRejections (node:internal/process/task_queues:82:21)

  3) Main Thread Detection
       should detect isMainThread in child_process:
     Uncaught TypeError [ERR_INVALID_ARG_TYPE]: The "options" argument must be of type object. Received an instance of Array
      at ChildProcess.target.send (node:internal/child_process:716:7)
      at Array.forEach (<anonymous>)
      at dispatchQueuedRequests (src/WorkerHandler.js:262:21)
      at ChildProcess.<anonymous> (src/WorkerHandler.js:221:7)
      at ChildProcess.emit (node:events:378:20)
      at emit (node:internal/child_process:920:12)
      at processTicksAndRejections (node:internal/process/task_queues:82:21)
```

Refs: https://github.com/nodejs/node/pull/37028/files#r566771229

PR #309, opened 2021-06-16
fix node 16 for v2 

mansona:
This PR backports the fix from https://github.com/josdejong/workerpool/pull/230 into v2 of workerpool. 

I also backported the change to Github CI and added Node 12 14 and 16 to the test matrix in a separate commit so you can see Node 16 fail in one commit and then see it being fixed in the next 👍

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,12 @@ jobs:
             install-command: npm ci
           - node-version: 10.x
             install-command: npm ci
+          - node-version: 12.x
+            install-command: npm ci
+          - node-version: 14.x
+            install-command: npm ci
+          - node-version: 16.x
+            install-command: npm ci
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


josdejong:
Thanks Chris, looks good 👍

josdejong:
I've just published v2.3.4. 

It wasn't very straightforward: the tests where passing but the build was broken, I had to switch to a very old version of node.js, and also changed the fix to ES5 code (see 6bc5913)

josdejong:
hm, now v2.3.4 is marked as latest on npm. Will publish a bug fix release on v6 to make that latest again...

josdejong:
ok fixed now by publishing v6.1.5 😅

mansona:
woops! thanks for sorting all that! 

Also sorry I didn't notice that there was a build, I should have checked that 🙈

josdejong:
You're welcome. Only if you try to publish a 4 year old library version again you realize how fast the ecosystem is moving...

In the latest versions of `workerpool`, the build is actually executed when testing, and there are tests to validate whether the generated library bundles actually do work.

Issue #322, opened 2021-09-15
Pass options to workers


FallingSnow:
Can you please add the ability to pass options to the workers created?
Worker syntax is `var myWorker = new Worker(aURL, options);`.

Need this to pass `type: 'module'` to the worker.

https://developer.mozilla.org/en-US/docs/Web/API/Worker/Worker

josdejong:
Thanks for picking this up @binvb in #329 👍

PR #329, opened 2021-11-25
feat(web worker): add init options #322

binvb:
reference https://github.com/josdejong/workerpool/issues/322

### DIFF
--- a/src/WorkerHandler.js
+++ b/src/WorkerHandler.js
@@ -70,7 +70,7 @@ function getDefaultWorker() {
 function setupWorker(script, options) {
   if (options.workerType === 'web') { // browser only
     ensureWebWorker();
-    return setupBrowserWorker(script, Worker);
+    return setupBrowserWorker(script, options, Worker);
   } else if (options.workerType === 'thread') { // node.js only
     WorkerThreads = ensureWorkerThreads();
     return setupWorkerThreadWorker(script, WorkerThreads);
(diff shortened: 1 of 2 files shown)


binvb:
oop, still need disting options

josdejong:
Thanks for your PR @binvb , this will be a useful extension! Sorry for my late reply.

I think the PR needs a bit of refinement. Feedbacks:

1. I think that the code of `resolveBrowserOptions` can be simplified a bit, it loops over all keys of `opts` and for each key it searches the array `workderOptArr`. Instead, you can loop over `workderOptArr` and copy the property when it is defined.
2. can you document the new options in the readme?
3. it would be nice to unit test this behavior, but I'm not sure how to do that for real since we don't have a test setup that can run in the browser. Do you have any ideas? At least it is possible to validate that only the three allowed options are copied and passed to setupBrowserWorker via setupWorker. Maybe it's also interesting to create a minimal  [example](https://github.com/josdejong/workerpool/tree/master/examples/browser) demonstrating the behavior so we can at least manually validate that it works for real. What do you think?

josdejong:
I was just thinking: it is probably a good idea to not pick these three properties from the generic `options` object (which contains options for the workerpool), but create a separate, nested object just like `forkArgs` and `forkOpts` (for nodejs processes). That way, we don't have to pick the properties but can pass the whole object as-is and won't mix these two different "levels" of options.

josdejong:
@binvb are you still interested in getting this PR finished? It would be a useful addition. 

If not, is anyone else interested in helping out here?

DonatJR:
@josdejong I would be interested in getting this finished. Are you okay with me creating a new pull request once I have the changes ready?

josdejong:
Thanks @DonatJR , a new PR would be very welcome!

josdejong:
Closing this PR now in favor of #400, which is published now in `v6.5.0`.

PR #335, opened 2022-01-19
Ensure pools are terminated after each test

forty:
As discussed in a previous PR, this avoids hanging the test suite in case terminate() is not called (which may happen on some failure case)


josdejong:
Nice, thanks Quentin! I like the `after(() => { ... }` inside the `createPool`, that even prevents needing global variables for the created pool, I didn't know that was possible this way 👌

PR #350, opened 2022-07-18
Put the copyright owner into license file.

MtTsai:
It's good to fill the owner information into license file.


josdejong:
o wow, I never noticed. Thanks!

Issue #370, opened 2023-01-05
Send event / message to running Worker?


flipswitchingmonkey:
The backchannel from the worker back to the main thread via emit works fine, but are there ports set up the other way around (sending a message into the worker?)

Reason: I need to send updated data to a running process inside a worker.

flipswitchingmonkey:
I guess it was the second part of this PR that was never implemented? :( https://github.com/josdejong/workerpool/pull/210

josdejong:
Thanks for your suggestion Michael. Yes there was a start with something like this in #210, but we still have to think through a good API for this.

Help would be welcome!

flipswitchingmonkey:
Hm. Just spitballing here... how about a new callback, similar to on() events. Something like
```ts
let executionId: number | undefined;
pool.exec('eventExample', [], {
  onExecutionStart: (id) => {
     executionId = id;
  },
  on: function (payload) {
   //...
  },
});

...

if (executionId) {
  pool.getWorkerInstance(executionId).emit('Take this!', {data: {...}});
}
```

josdejong:
Yes that could work, though it involves an indirect solution via an identifier.

How about extending the returned Promise-like instance:

```js
const handler = pool.exec('fibonacci', [10], {
  on: function (payload) { ... }
}

// the returned promise-like `handler` actually is an extended version of Promise, 
// having methods .then(callback), .catch(callback), .cancel(), .timeout(delay), and .emit(payload)
promise.emit('Take this!')

const result = await handler
```

To prevent confusion, we should maybe rename `Promise` to `ExecutionHandler` or something like that.

flipswitchingmonkey:
That'd be much nicer of course!

Also, it'd be nice if the messages would be queued up as well, in case the worker has not started yet (and alternatively, the promise should also return the current state of the task (queued, running, finished,...) so messages don't get lost)

Adding this would make the pool so much more useful for controlling long running tasks!

josdejong:
Yes, it will be nice to be able to have two way messaging between the main process and the worker.

I'm not sure if it is possible to queue messages upfront: upfront you do not know at which worker a task will be assigned, only as soon as the task started you can send something to the worker where the task is running I think?

Maybe you can share your use-case to better understand what a good API would be to cater for it?

flipswitchingmonkey:
It'd be enough if the promise returned the state so that only in case it is running, a message could be sent.

[...]

PR #372, opened 2023-01-17
adds worker-side on(message) events

flipswitchingmonkey:
This PR is related to https://github.com/josdejong/workerpool/issues/370

A rather small change that adds a second parameter to `workerpool.worker()` function, which accepts an object where each key is the receiver's name of an event, plus a payload. 

It also adds an emit() function to the Promise object which will send a postMessage() call to the worker instance.

These changes should be non-breaking.

Example:
workers/myworker.js
```js
workerpool.worker({
  fibonacci: fibonacci,
}, {
  callme: function (payload) {
    makeThingsHappen(payload);
  }
});
```

And to call the event:
```
const pool = new Pool(__dirname + "/workers/myworker.js");

// main difference here is that the handler, a Promise, is picked up from the exec() command first
const handler = pool.exec('fibonacci', [15]);

// now continue the chaining
handler
  .then(function (result) {
    assert.strictEqual(result, 610);
  })
  .catch(function (err) {
  })
  .then(function () {
    pool.terminate(); // terminate all workers when done
  })
  
// the Promise now has a function Promise.emit(name, paramerers) 
handler.emit('callme', {say: 'O hai!'});
```


test/CallMethod.test.js
```diff
@@ -0,0 +1,42 @@
...
+  .catch(function (err) {
+    assert.strictEqual(err.toString().includes('exitCode: `99`'), true);
+  })
+  .then(function () {
+    pool.terminate(); // terminate all workers when done
+    done();
+  })
+  .catch(done);
+  
+  handler.emit('killme', 99); // forces the worker to exit with exitcode 99
```
josdejong, line 41: I have the feeling that the unit test testing the behavior can be simplified a bit, rather than indirectly testing for the worker killing itself as a result of receiving a message. How about changing it to something more direct like:

```js
const handler = pool.exec('getEmittedMessage', [])
  .then(function (result) {
    assert.strictEqual(result, 'hello world')
  })
    
handler.emit('myCustomEvent', 'hello world')
```

What do you think?

src/Promise.js
```diff
@@ -141,6 +144,22 @@ function Promise(handler, parent) {
     return me;
   };
 
+/**
+   * Calls a defined worker method with optional parameters
+   * @param {string} methodName Name of the method to be called as defined in the workerpool.worker() function
+   * @param {Array} [methodParameters] Array with optional parameters
+   */
+this.emit = function (methodName, methodParameters) {
+  if (this.worker?.worker && methodName) {
```
josdejong, line 153: Right now, an emitted event will be silently ignored when a worker is not yet attached (i.e. the task is queued). I think it would be better to throw an exception, or queue the message and execute. Or maybe return a boolean indicating whether the message was sent or not. What do you think?

src/Promise.js
```diff
@@ -141,6 +144,22 @@ function Promise(handler, parent) {
     return me;
   };
 
+/**
+   * Calls a defined worker method with optional parameters
+   * @param {string} methodName Name of the method to be called as defined in the workerpool.worker() function
+   * @param {Array} [methodParameters] Array with optional parameters
+   */
+this.emit = function (methodName, methodParameters) {
```
josdejong, line 152: Can you rename this to `function (eventName, eventParams) {` ?

src/index.js
```diff
@@ -16,9 +16,10 @@ exports.pool = function pool(script, options) {
  * Create a worker and optionally register a set of methods to the worker.
  * @param {Object} [methods]
  */
-exports.worker = function worker(methods) {
+exports.worker = function worker(methods, events) {
   var worker = require('./worker');
   worker.add(methods);
+  if (events) worker.addOnEvents(events);
```
josdejong, line 22: Personal preference: I find the name `addOnEvents` a bit confusing. How about naming it `registerEvents`  or `registerEventListeners` for example?

josdejong (changes requested):
Thanks, this looks really promising @flipswitchingmonkey! I like the `workerpool.worker(methods, events)` API, that makes sense.

I made a couple of inline comments, can you have a look at those?

And lastly, we should document the new feature in the README.md.

blordpluto:
The functionality here is crucial for my team to use this package. This PR appears to have been abandoned, and now unsurprisingly conflicts greatly. @josdejong what would be your level of interest in re-approaching this?

josdejong:
@blordpluto thanks for your input. I would love to see this feature finished. The PR was under review but not yet there. Would you be interested in creating a new PR implementing this feature?

PR #377, opened 2023-02-19
feat: support cleanup on termination in worker context

### LINKED ISSUE #353: Allow main thread to cleanly cancel workers
The `cancel` method of the worker promise sends a signal that kills the worker thread entirely, leaving any resources used in the thread in a state of dismay. 

For example, my worker starts processes of its own and use and produces output files, etc. If this task should be cancelled, I would like my worker to properly terminate those processes and make sure the files are in a usable state (close json objects, etc). Now, if the worker task is canceled, all those resources are left as they last were and pollute the system.

Is it possible to send a softer cancel signal to the worker instead of the current kill? Or is there any mechanism by which the main thread could send some message that the worker can listen to and avoid using that cancel signal?

Michsior14:
Adds termination handler in a worker context to allow releasing resources.

Tests will be added after feedback (to confirm interfaces/idea).

Resolves #353


josdejong:
Thanks, I'll review your PR tomorrow. I see there are no tests added. Would it be possible to unit test the new cleanup functionality?

Michsior14:
> Thanks, I'll review your PR tomorrow. I see there are no tests added. Would it be possible to unit test the new cleanup functionality?

As in the PR description I wanted to first confirm this is the way to go ;)

josdejong:
ah sorry, makes sense 😅

josdejong:
This looks really good @Michsior14, exactly what we need 👌 . If you can add tests I think all is ready to merge this PR.

Michsior14:
@josdejong I've added working tests using `MessageChannel` since I had no other idea how to tackle this.

josdejong:
Thanks! It is indeed not so trivial to test whether `onTerminateWorker` since it is shutting down the worker. Sounds like a good solution to use `MessageChannel`! I'll now merge your PR and publish a new version of `workerpool` at npm.

josdejong:
Published now in `v6.4.0`, thanks again!

PR #388, opened 2023-04-24
Promise: add .finally()

### LINKED ISSUE #169: Promise.prototype.finally
`Promise.prototype.finally()` was added in Node v10 and is also in most major browsers now. It would be great to have that function available from `pool.exec`. 

Some references:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally
https://github.com/tc39/proposal-promise-finally/blob/fd934c0b42d59bf8d9446e737ba14d50a9067216/polyfill.js

wmertens:
Fixes #169 

Thanks to https://stackoverflow.com/a/53327815/124416


josdejong:
Thanks Wout!

Interesting to see that the `finally` propagates both the result or error, I wasn't aware of that.

Two feedbacks:
- I'm wondering if there is need to document this method. I think not, since it's so standard. What do you think?
- Can you write unit tests for this method?

joshLong145:
@josdejong @wmertens
This will be a useful addition until there can be a migration to the native Promise library. I have added some tests in a new branch to confirm matching functionality with the native `finally` implementation.

https://github.com/joshLong145/workerpool/tree/feat/add-promise-finally

josdejong:
Thanks for the update Wout, it has been a while 😅 .

Two remarks: 

1. Can you describe the new `.finally()` method in the README.md? There is a list with all methods there that can be extended.
2. I think the unit tests need to be extended a bit: when I remove the `.then` and `.catch` from the two tests, the test still succeeds. So, I think the test should check whether the `.then` and `.catch` have been trigered before calling `done`. It is valuable at least to verify the order in which the `.then`, `.catch`, and `.finally` are triggered so these two tests make sense to me.
3. Similarly, the `finally` method neatly propagates either the success or failure, but this is not yet captured in a unit test. Can you add two unit tests for that?

src/Promise.js
```diff
@@ -196,6 +196,20 @@ Promise.prototype.always = function (fn) {
...
 };
 
+/**
+ * Execute given callback when the promise either resolves or rejects.
+ * Same semantics as Node's Promise.finally()
+ * @param {Function} fn
+ * @returns {Promise} promise
+ */
+Promise.prototype.finally = function (fn) {
+  const onFinally = callback => Promise.resolve(fn()).then(callback);
```
joshLong145, line 206: @wmertens This will not work with the `Promise` implementation used here as `Promise.resolve` is not implemented. I think we can do something like

```js
const onFinally = (callback) => {
  const wrapper = Promise.defer();
  wrapper.promise.then(callback);
  wrapper.resolve(fn());
  return wrapper.promise;
}
```

wmertens: Ah indeed - I gave you access to my fork, feel free to add commits to this branch, unfortunately I'm low on time for the moment 🙏

joshLong145: I ended up with an implementation
```javascript
Promise.prototype.finally = function (fn) {
  var me = this;
  const res = function() {
    return me;
  }
  const final = () => {
    const wrapper = Promise.defer();
    wrapper.promise.then(res);
    wrapper.resolve(fn());
    return wrapper.promise;
  };

  return this.then(final, final);
}
```

This allows `finally` to execute if the promises is `resolved` or `rejected` and allows for the promise chain to continue. If you are okay with this I can push to the branch along with some changes to unit tests.

joshLong145:
@josdejong @wmertens

PR has been updated with
- new `finally` implementation
- method documentation in README
- test case additions to `Promise.test.js`

src/Promise.js
```diff
@@ -210,6 +210,27 @@ Promise.prototype.always = function (fn) {
...
+  * @returns {Promise} promise
+  */
+Promise.prototype.finally = function (fn) {
+  var me = this;
+  const res = function() {
+    return me;
+  }
+  const final = function() {
+    const wrapper = Promise.defer();
+    wrapper.promise.then(res);
```
wmertens, line 226: Actually, what does this line do? Is there some side effect?

joshLong145: This chains the current Promise as the return from the newly created promise as to keep the promise chain in tact. Without this a call to finally would result in the promise chain ending.

wmertens: Shouldn't that then be `wrapper.promise = wrapper.promise.then(res)`?

joshLong145: Good point! I went ahead and updated the imlpementation to no longer use `Promise.defer` and to just new up a `Promise` instance as it simplified the implementation since we always `resolve` the promise.

I also removed the test checking for a value returned from `finally` as I was mistaken and the spec does not specify this.

test/Promise.test.js
```diff
@@ -211,6 +211,74 @@ describe ('Promise', function () {
...
+        finallyRan = true;
+      }).then(function() {
+        assert.equal(finallyRan, true);
+      });
+
+      assert.strictEqual(p.resolved, true);
+      assert.strictEqual(p.rejected, false);
+      assert.strictEqual(p.pending, false);
+
+      done();
```
josdejong, line 234: I think the `done()` in this unit test should be _inside_ the last `.then()`, like:

```js
.then(function() {
  assert.equal(finallyRan, true);
  done();
});
```

It works right now, but in general, a promise runs async so the `done` would be called before the `assert.equal(finallyRan, true)`.

test/Promise.test.js
```diff
@@ -211,6 +211,74 @@ describe ('Promise', function () {
...
+      var p = new Promise(function(_resolve, reject) {
+        reject(new Error('An error has occured')); 
+      }).catch(function(_err) {
+        isFullfilled = true;
+        return err;
+      }).finally(function(value) {
+        assert.ok(isFullfilled, "should call finally after reject");
+        assert.equal(value, undefined);
+        finallyRan = true;
+      });
```
josdejong, line 249: Can you add the test whether `finallyRan` is `true` here too, like in the other two tests, and call `done` inside a last `.then()`?

test/Promise.test.js
```diff
@@ -211,6 +211,74 @@ describe ('Promise', function () {
...
+      var p = new Promise(function(resolve, reject) {
+        resolve(); 
+      }).then(function () {
+        isFullfilled = true; 
+      }).finally(function(value) {
+        assert.ok(isFullfilled, "should call finally after resolve");
+        assert.equal(value, undefined);
+        finallyRan = true;
+      }).then(function() {
+        assert.equal(finallyRan, true);
```
josdejong, line 270: Maybe move this `assert.equal(finallyRan, true)` into the last `return p.then(...)` to have all these final tests in a single place?

josdejong (changes requested):
Thanks @joshLong145 for the updates! This looks good.

I made three small comments in the unit tests, can you have a look at those?

test/Promise.test.js
joshLong145: Yes this makes sense. I was following the other tests in `Promise.test.js` I can refactor this to run in the promise chain.

test/Promise.test.js
joshLong145: Yes good catch!

joshLong145:
> Thanks @joshLong145 for the updates! This looks good.
> 
> I made three small comments in the unit tests, can you have a look at those?

Thanks @josdejong 
Made those changes

josdejong (approved):

josdejong:
Thanks for the updates Josh!

josdejong:
Published now in `v9.2.0`, thanks again!

Issue #387, opened 2023-03-30
Possible bugs in the code.


iSuslov:
While I'm rewriting the code to Typescript, I found two things that I would rather ask here since it looks like bugs.  

1. Lets take a look here:

https://github.com/josdejong/workerpool/blob/8219f7deaf535310cb8c3f273cda8e047959502c/src/WorkerHandler.js#L483-L494

On the line 490 there is a assignment `resolver.promise.timeout = timeout;`. Doc of the method says timeout param is a `number`, but `Promise` object has a method called `timeout`. Seems like an attempt to override a method with a number. Task object however has a property `timeout:number`. Luckily it seems nowhere in the code this method `WorkerHandler.prototype.terminateAndNotify` gets called with the second argument. 

Is it a bug in abandoned code?

2. Second one. Lets take a look at how `TERMINATE_METHOD_ID` is used.

https://github.com/josdejong/workerpool/blob/8219f7deaf535310cb8c3f273cda8e047959502c/src/WorkerHandler.js#L11

https://github.com/josdejong/workerpool/blob/8219f7deaf535310cb8c3f273cda8e047959502c/src/WorkerHandler.js#L450-L454

There is no problem when it gets directly sent to a worker, however when it gets added to the `requestQueue` there is a problem, because this is how `requestQueue` is being processed:

https://github.com/josdejong/workerpool/blob/8219f7deaf535310cb8c3f273cda8e047959502c/src/WorkerHandler.js#L272-L278

Looks like instead of `TERMINATE_METHOD_ID` undefined gets sent. Correct me if I'm wrong please.

Michsior14:
Both bugs have been resolved :)

PR #394, opened 2023-08-17
fix: worker termination before it's ready

Michsior14:
Fixes the second issue mentioned in possible bugs thread (#387).


josdejong:
Good find, thanks Michał!

TypeScript would have helped to prevent a bug like this I guess 🤔.

josdejong:
Ready to merge?

josdejong:
Your fix is published now in `v6.4.1`, thanks again!

Michsior14:
> Good find, thanks Michał!
> 
> TypeScript would have helped to prevent a bug like this I guess 🤔.

Yeah, maybe I will find some time to continue what I started some time ago... but no promises :)

PR #395, opened 2023-08-17
fix: properly timeout on worker termination

Michsior14:
Fixes the termination bug mentioned at #387.


josdejong:
Thanks @Michsior14 this fix makes sense indeed!

josdejong:
Fix published now in `v6.4.2`

PR #400, opened 2023-09-11
Support for passing options to web worker constructor

DonatJR:
Add the ability to pass options to web worker constructor: see #322 for initial request and #329 for previous PR.

I am not sure about typing the `WorkerOptions` in `src/types.js`, since they are taken straight from the spec and would need to be updated if the spec ever changes. Should I replace them with a generic `*` like `forkArgs` and `forkOpts`?


josdejong:
Thanks Jonas, this looks straightforward and well documented, including a clear example!

I like that you worked out the types for `WorkerOptions`. I think they will not change quickly, and if so, we can always update.

DonatJR:
@josdejong Thank _you_ for merging and publishing so fast, cheers! 🎉

PR #408, opened 2023-10-25
Example vite and webpack5

KonghaYao:
After updating to version 7.0.0, I created a simple example for both dev and production modes using Vite and Webpack 5. The example includes some adaptation code for Vite and Webpack, which I believe will be helpful for those who are not very familiar with web workers.

### DIFF
--- a/examples/webpack5/src/vite-env.d.ts
+++ b/examples/webpack5/src/vite-env.d.ts
@@ -1,2 +1,0 @@
-/// <reference types="vite/client" />
-
--- a/examples/webpack5/src/vite-env.d.ts
+++ b/examples/webpack5/src/vite-env.d.ts
@@ -1,0 +1,2 @@
+/// <reference types="vite/client" />
+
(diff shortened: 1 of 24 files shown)


examples/webpack5/src/vite-env.d.ts
josdejong, line 1: Can you remove this vite file from the webpack5 example? I think it would be confusing :)

josdejong (commented):
Thanks, this is great and indeed very helpful I think!

examples/webpack5/src/vite-env.d.ts
KonghaYao: Oops! I forgot to delete this file! It has been modified now. 😂

josdejong: 💪

KonghaYao:
You're welcome. Workerpool has been very helpful in my project, and I also want to help it get better and better.

josdejong:
Thanks, appreciated!

PR #409, opened 2023-11-16
Fix a typo in README.md

senicko:
It looks like the first "can" in "can can" was meant to be "and"


josdejong:
Thanks Sebastian! I'm quite sure I didn't mean [can-can](https://en.wikipedia.org/wiki/Can-can) when I wrote this sentence 😉, and I think you're right.

PR #413, opened 2023-12-09
Add an example for an inline worker with esbuild

tamuratak:
This PR adds an example for an inline worker with esbuild. The build script is based on @manzt's https://gist.github.com/manzt/689e4937f5ae998c56af72efc9217ef0.

We could say this is more simple rather than `examples/embeddedWorker`.


josdejong:
Thanks! 

All in all I think both `embeddedWorker` and this new `esbuild` example are relatively complex. I like the `vite` example better 😉.

Can you please rename the example folder to just "esbuild", and fix the indentation of the new files and make those consistent?

josdejong:
Thanks for the update! Merging the example now.

PR #414, opened 2023-12-09
Fix examples/embeddedWorker

tamuratak:
Fix `examples/embeddedWorker`. The current one doesn't work with webpack 5.


josdejong:
Thanks Takashi! Too bad there is more config needed now to make this work with Webpack.

PR #415, opened 2023-12-11
Generate types from JSDoc comments with TypeScript compiler.

### LINKED ISSUE #334: Move typescript definitions of @types/workerpool into the library and bring the definitions up to date
Having the type definitions in the repo itself will make it easier to keep the definitions on par with the API.

Anyone able to help out here?

tamuratak:
Generate types from JSDoc comments with TypeScript compiler. Close #334. 

TypeScript compiler can generate `.d.ts` files from JavaScript files. See

- https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html

To make it work, we need minor modifications on `src/index.js` and `types.js`.

Run `npm run types` to generate types.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,13 +27,14 @@
     "README.md"
   ],
   "scripts": {
-    "build": "rollup -c rollup.config.mjs",
+    "build": "rollup -c rollup.config.mjs && npm run build:types",
+    "build:types": "tsc -p .",
     "watch": "rollup -c rollup.config.mjs -w",
-    "test": "npm run build && mocha test",
+    "test": "npm run build && mocha test && npm run test:types",
+    "test:types": "tsc -p test/types",
     "test:debug": "npm run build && mocha debug test",
     "coverage": "npm run build && c8 mocha && c8 report --reporter=html && echo Coverage report is available at ./coverage/index.html",
-    "prepublishOnly": "npm run test",
-    "types": "tsc -p ."
+    "prepublishOnly": "npm run test"
   },
   "devDependencies": {
     "@types/node": "20.10.4",
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
   ],
   "main": "src/index.js",
   "browser": "dist/workerpool.js",
+  "types": "types/index.d.ts",
   "files": [
     "dist",
     "src",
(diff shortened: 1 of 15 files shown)


src/Pool.js
```diff
@@ -56,6 +75,7 @@ function Pool(script, options) {
     this._ensureMinWorkers();
   }
 
+  /** @private */
   this._boundNext = this._next.bind(this);
```
josdejong, line 79: I think the `@private` and `@protected` annotations on methods are bit mixed (that already was the case). Shall we make all internal methods in `Pool.js` `@private` (they are prefixed with a `_`)?

package.json
```diff
@@ -31,9 +32,11 @@
     "test": "npm run build && mocha test",
     "test:debug": "npm run build && mocha debug test",
     "coverage": "npm run build && c8 mocha && c8 report --reporter=html && echo Coverage report is available at ./coverage/index.html",
-    "prepublishOnly": "npm run test"
+    "prepublishOnly": "npm run test",
+    "types": "tsc -p ."
```
josdejong, line 36: Generating the types by running `npm run types` should automatically be done before publishing. Can you rename the script `build:types` and add running it to `prepublishOnly`?

josdejong: Or we can extend the `build` script to also generate the types, we have to test them afterwards too.

package.json
```diff
@@ -31,9 +32,11 @@
     "test": "npm run build && mocha test",
```
josdejong, line 32: We need to add a test to verify that the types are generated correctly, like a TS file `testTypes.ts` that spins up a simple workerpool, and that we use to verify that it compiles without TS errors. What do you think?

josdejong (commented):
This is a very nice idea Takashi, thanks a lot! In the end I would like to refactor the code to TS and ESM, but this is a great first step in that direction. 

Reminder to myself: once merged, I'll publish it as a breaking change since the types may differ slightly from `@types/workerpool`.

tamuratak:
We have to tweak `./src/Promise` to make the generated types compatible with types/workerpool. Please check whether the change is accepetable.

josdejong (approved):
👌  Thanks for the updates, this looks good to me.

josdejong:
Published now as `v9.0.0`. Thank again Takashi!

PR #416, opened 2023-12-18
Export more types

joshkel:
This allows code using workerpool to easily reference pool types. For example, I could write code similar to the following in workerpool 8, but in workerpool 9, I could not find an easy way to access these types, since they're just imports within the generated `index.d.ts`.

```ts
import { pool, type WorkerPoolOptions, type WorkerPool } from 'workerpool';

function getOptions(): WorkerPoolOptions { /* ... */ };
function makeMyPool(): WorkerPool {
  const myPool = pool(getOptions());
  // initialize
  return myPool;
}
```

(workerpool 9's build-in types also rename `WorkerPool` to `Pool`, but that's an easy change to make in calling code.)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,16 +1,20 @@
 const {platform, isMainThread, cpus} = require('./environment');
 
+/** @typedef {import("./Pool")} Pool */
+/** @typedef {import("./types.js").WorkerPoolOptions} WorkerPoolOptions */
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks for the improvement @joshkel, this makes sense!

josdejong:
Published now in `v9.0.1`. Please let me know if you encounter more issues with the types!

joshkel:
Works great. Thanks, @josdejong!

PR #417, opened 2023-12-18
Fix types

tamuratak:
This PR is a minor fix for types.

- export the Proxy type.
- should use the workerpool's Promise type.


josdejong:
Thanks!

josdejong:
Fix is published now in `v9.0.2`.

PR #419, opened 2023-12-19
Unwrap promise types

joshkel:
A promise's `then` or `catch` handler is allowed to return a promise itself; that promise is automatically awaited before invoking the next `then` or `catch` in the chain.  However, workerpool's types did not reflect this: as a result, promise chains could end up with types like `Promise<Promise<number, Error>, any>`.


josdejong:
Good point, thanks!

josdejong:
Published now in `v9.0.3`

PR #421, opened 2023-12-21
Improve the logic of isNode().

### LINKED ISSUE #320: Improve logic to determina whether in a browser or nodejs environment
When using this with jspm.org and in the browser, the library thinks it's in node.

This is because the library is webpacked before distribution. Webpack adds
```
import process from "process";
```
to the top of the environment.js file. JSPM then goes through each dependency and tries to pull it from npm. And unfortunately there is an NPM module named `process` which it pulls. That module exports `process` to the browser. So in the browser you see `process.versions.node`.
Screenshot from 2021-09-12 15-14-29

At least that's how I understand it so far.

I think something more than just process should be relied on, or not at all. Perhaps `self.naviagator`?
It has good support: https://caniuse.com/mdn-api_navigator

tamuratak:
Improve the logic of `isNode()`. Close #320.

The trick of `'[object process]'` seems to be widely used.

- https://github.com/search?q=%22%5Bobject+process%5D%22&type=code
- https://github.com/mozilla/pdf.js/blob/7ea0e40e588864cd938d1836ec61f1928d3877d3/src/shared/util.js#L24

Now, since the `process.versions` is empty on `jspm-core`, you could close the original issue without merging this PR.

- https://github.com/jspm/jspm-core/blob/7af7d7413f472305d08d0d78ec3d1f15588be50a/nodelibs/browser/process.js#L78

### DIFF
--- a/src/environment.js
+++ b/src/environment.js
@@ -1,10 +1,12 @@
 
 // source: https://github.com/flexdinesh/browser-or-node
+// source: https://github.com/mozilla/pdf.js/blob/7ea0e40e588864cd938d1836ec61f1928d3877d3/src/shared/util.js#L24
 var isNode = function (nodeProcess) {
   return (
     typeof nodeProcess !== 'undefined' &&
     nodeProcess.versions != null &&
-    nodeProcess.versions.node != null
+    nodeProcess.versions.node != null &&
+    nodeProcess + '' === '[object process]'
   );
 }
 module.exports.isNode = isNode
--- a/test/environment.test.js
+++ b/test/environment.test.js
@@ -8,10 +8,11 @@ describe('Environment Detection', function () {
     assert.strictEqual(isNode(undefined), false)
     assert.strictEqual(isNode({}), false)
     assert.strictEqual(isNode({ versions: {} }), false)
+    assert.strictEqual(isNode({ versions: { node: '10.0.0' } }), false)
   })
 
   it('is platform assigned to be node', function () {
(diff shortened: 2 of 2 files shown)


josdejong:
Thanks, this approach makes sense. How about removing the logic that checks `.node.versions`? I think that is redundant now. It can become just `nodeProcess + "" === "[object process]"`, right ? Or `String(nodeProcess) === "[object process]"` to make it a bit more explicit.
```

> Now, since the `process.versions` is empty on `jspm-core`, you could close the original issue without merging this PR.

Do you mean that #320 _is_ already fixed without this PR?

tamuratak:
> How about removing the logic that checks .node.versions?

It would break `workpool` on the Electron renderer process, which has the built-in `process` although we should consider it as a `browser` platform.

- https://www.electronjs.org/docs/latest/api/process

> Do you mean that #320 _is_ already fixed without this PR?

Yes. This PR would be good for another polyfill library that will appear in the future.

josdejong:
Ah, of course. In the [`isNodeJs`](https://github.com/mozilla/pdf.js/blob/7ea0e40e588864cd938d1836ec61f1928d3877d3/src/shared/util.js#L24) function of pdf.js that you linked to there are also a couple of extra conditions to distinguish from Electron and NW I see. Ok let's keep the logic that we already had as is 👍 .

josdejong:
Published now in `v9.0.4`, thanks again. I wish you nice holidays!

Issue #438, opened 2024-04-02
Webworker implementation does not handle on message callbacks.


t0oF-azpn:
I tried to reproduce the example of sending back data from the worker to the pool (from the doc https://github.com/josdejong/workerpool#events), in a browser environment (ie. with web workers):

https://codesandbox.io/p/sandbox/workerpool-in-a-react-project-forked-dgjxxc

The example demonstrates 2 issues:
- `pool.exec("someFunctionName")` does fail in a browser environment 
- `pool.exec(someFunction)` does perform the computation but fails to send the web-worker message to the pool.

josdejong:
Thanks for reporting and implementing a fix in #439.

josdejong:
Fixed in `v9.1.1`

PR #439, opened 2024-04-02
fix: pool.exec(<function>) fails to emit message event to the pool.

t0oF-azpn:
When using worker-pool in a browser environment, worker function registration does not work. Only `pool.exec(<function>)` can be run, but does not handle the on message callbacks.

This fix proposes to send the `pool.exec` options to the execution script.

Referenced doc: https://github.com/josdejong/workerpool#events
Playground: https://codesandbox.io/p/sandbox/workerpool-in-a-react-project-forked-dgjxxc

Possible fix for issue #438

### DIFF
--- a/src/Pool.js
+++ b/src/Pool.js
@@ -166,7 +166,7 @@ Pool.prototype.exec = function (method, params, options) {
   }
   else if (typeof method === 'function') {
     // send stringified function and function arguments to worker
-    return this.exec('run', [String(method), params]);
+    return this.exec('run', [String(method), params], options);
   }
   else {
     throw new TypeError('Function or string expected as argument "method"');


t0oF-azpn:
> Thanks @t0oF-azpn for reporting this issue and working out a fix!
> 
> Would it be possible to write a unit test to ensure we can't get any regressions in the future?

I tried first to add/update the unit tests. But I didn't manage to make it run.
In this particular case, we need to be in browser context. I tried to force to use web workers (`var pool = createPool({ workerType: 'web' });`), but this fails at worker creation in `ensureWebWorker()` function.

If there is any way to set the test to web browser context, I would take it ;)

josdejong:
OK let's just merge this PR as is, ok? The fix makes sense. 

I do think though that referring to `worker` in a standalone function is a bit of a tricky case: you're relying on an undocumented global variable exposed by a library. The neat solution would be to create a dedicated worker where you import `worker` explicitly.

t0oF-azpn:
I'm fine with merging This PR as is !
I don't have the rights to do it, though ;)

josdejong:
👍 I'll merge and publish your PR first half of next week.

josdejong:
I found a moment, just published `v9.1.1` containing your fix. Thanks again!

PR #442, opened 2024-04-16
add execDirect

frankie-zeng:
allow multi task run in worker at same time
```
var workerpool = require('../../');
var ac
function start() {
  ac=new AbortController();
  return new Promise(function (resolve, reject) {
    ac.signal.onabort = () => {
        resolve('aborted');
    };
});  
}
function stop() {
  ac.abort();
}

// create a worker and register some functions
workerpool.worker({
    start: start,
    stop:stop
});
```
```
pool.execDirect('start',[]).then((val)=>{
      assert.strictEqual(val,'aborted');
      done();
    }).catch(done)
    pool.execDirect('stop',[])
```


josdejong:
Thanks for your PR. Can you elaborate on what the PR does? Is it about adding support for an `AbortController` like in #441?

I see you deleted test files `Pool.test.js` and `Promis.test.js`, that should not be done. These test files are necessary to ensure the behavior of the library is as we want.

josdejong:
?

frankie-zeng:
> ?

I will fix the code and reopen a new PR. thanks.

PR #448, opened 2024-06-11
Feat/abort listener

joshLong145:
Adds an `abortListener` to workers which allow for cleanup of async tasks which can be run as a cleanup operation to allow workers to be reused if a task timeout or cancellation occurs. 

connects PR #441


test/Pool.test.js
```diff
@@ -1315,6 +1315,22 @@ describe('Pool', function () {
     });
   });
 
+  it('should not terminate worker if abort listener is defined', function () {
```
josdejong, line 1318: I think we can use some additional unit tests:

1. test that checks that the second task does not trigger the abort listener of the previous task (ie. verify that the listener is cleaned up)
2. test the ability to attach multiple abort listeners (for example 2)
3. test the behavior when the abort listener never resolves
4. verify that you can use an abort listener in an offloaded function (not a dedicated worker)

Does that make sense? More ideas?

josdejong (changes requested):
Thanks a lot, this looks good! 

I made some inline comments, can you have a look at those?

src/worker.js
```diff
@@ -253,4 +270,7 @@ worker.emit = function (payload) {
 if (typeof exports !== 'undefined') {
   exports.add = worker.register;
   exports.emit = worker.emit;
+  exports.addAbortListener = function(listener) {
+    worker.onAbort = listener;
```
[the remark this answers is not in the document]
joshLong145: I think this makes sense, better to keep with standard listener implementation.

src/worker.js
```diff
@@ -121,16 +126,28 @@ worker.cleanupAndExit = function(code) {
...
+    worker.onAbort = undefined;
   }
 
-  var result = worker.terminationHandler(code);
-  if (isPromise(result)) {
-    result.then(_exit, _exit);
+  if (worker.onAbort) {
+    let handler = worker.onAbort();
+    if(isPromise(handler)) {
+      handler.then(_abort);
```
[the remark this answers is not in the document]
joshLong145: This makes sense I think we will need some `Promise.race` implementation to do it cleanly. Would we want this timeout to be global, or unique for each registered listener?

test/Pool.test.js
joshLong145: I think these are good surface area. Do we need to update any existing tests?

src/worker.js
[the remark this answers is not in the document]
josdejong: At first I was thinking about using `Promise.all`. It is neat though to resolve them one by one in order. I think there is no need to extract and pass `first` to the `reduce` method: if you omit the inital value, it will automatically pick the first element of the array as initial value. And I think you can write the body of the `reduce` method more compact, like `.reduce((prev, next) => prev.then(next))`. 

If one of the abort listeners throws an exepction, I think that we should fall back on running `_exit();`, what do you think?

src/worker.js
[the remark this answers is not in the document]
josdejong: Ok! I think the timeout should be a setting on the workerpool, like `minWorkers` and `maxWorker`, and not per task or listener.

test/Pool.test.js
josdejong: I think all existing tests can be in place: this will not be a breaking change in behavior, only a new feature, so we should not need to alter existing tests.

src/worker.js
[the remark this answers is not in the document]
joshLong145: I tried the reduce and I think its easier to do something like this

```js
    const promises = worker.abortListeners.map((listener) => listener());
    Promise.race([
          Promise.allSettled(promises).then(_abort),
          () => {
            return new Promise((resolve) => setTimeout(resolve, 300))  
          }
    ]).then(_abort, _exit);
```

So we just execute the listeners and wrapped promises with `Promise.allSettled` and just wait for all promises to fulfill before calling `_abort` or `_exit` respectfully,  I don't think sequential chaining is worth it as it will make handling the timeout a bit tricky.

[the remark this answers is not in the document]
josdejong: Agree, I think multiple listeners is a very small edge case already, let's keep it simple and resolve them in parallel 👍 

Your code example looks good indeed, with the `Promise.race`. And then the 300 ms can become a configurable option.

joshLong145:
After some testing of offloaded functions I think there are issues with referencing variables which are defined outside of the offloaded function scope

```js
    var workerpool = require('../');
    var pool = workerpool.pool();

    function asycTimeout() {
      return new Promise(function (resolve) {
        let timeout = setTimeout(function () {
            resolve();
        }, 5000);
        
        workerpool.addAbortListener(async function () {
            await new Promise((res, rej) => {
              setTimeout(res, 1000);
            });
            clearTimeout(timeout);
            resolve();
        });
      });
    };

    pool.exec(asycTimeout, [])
```

The above will error with `workerpool` is not defined. It does not seem like we can define an instance of `workerpool` within the global if using offloaded functions. I think the only way to support it would be to pass the registration handler to the function.

josdejong:
It sounds like a good idea to attach `this.addAbortListener` to the function itself rather than a "magic" global variable `workerpool.addAbortListener` 👍 . Can you make sure that it is possible to use it as follows?

```js
const addAbortListener = this.addAbortListener
// ...
addAbortListener(...)
```

joshLong145:
> EDIT: and then it makes sense to me to offer this as the only way to add an abort listener, for both offloaded and dedicated workers, right?

Yes this makes sense to me. can make the updates and implement the tests/examples now that we have this worked out.

Since we now can extend the function created for the worker task we can create a `worker` api with

- addEventListener
- emit

So for example

```js
this.worker.addEventListener
```

test/Pool.test.js
```diff
@@ -1315,6 +1315,22 @@ describe('Pool', function () {
...
+    var pool = createPool(__dirname + '/workers/cleanup-abort.js');
+    
+    pool.exec('asyncTimeout', [])
+    .timeout(200)
+    .catch((err) => {
+      assert(err instanceof Promise.TimeoutError);
+      // We should still have a busy worker since the worker should still be
+      // active and the promise chain for the timeout has yet to resolve
+      // which keeps the task active.
+      assert(pool.stats().busyWorkers === 1);
```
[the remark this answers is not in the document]
joshLong145: Emitting an event through `worker.emit` is not possible from the `onAbort` context since the task will be cleaned up if `timeout` or `cancel` is invoked by the time the `onAbort` handler is called.

src/worker.js
```diff
@@ -250,6 +280,21 @@ worker.emit = function (payload) {
   }
 };
 
+/**
+ * Builds an api for attaching to a method for
+ * worker functionality availble from within a task context
+ * @returns {Object}
+ */
+function _buildMethodWorkerApi() {
```
josdejong, line 288: I think instead of a function `_buildMethodWorkerApi`, we can create a new static object named for example `publicWorker` or so, alongside the `worker` object, and let it have the public methods `addAbortListener` and `emit`, since they will not change over time.

So we get something along the lines of:

```js
const worker = { ... }

const publicWorker = { 
  addAbortListener: worker.addAbortListener,
  emit: worker.emit
}

File: src/worker.js [hunk 6]
```diff
@@ -220,12 +247,15 @@ worker.register = function (methods, options) {
...
         worker.methods[name] = methods[name];
+        worker.methods[name].worker = _buildMethodWorkerApi();
       }
     }
   }
 
   if (options) {
     worker.terminationHandler = options.onTerminate;
+    // register listener timeout or default to 1 second
+    worker.abortListenerTimeout = options.abortListenerTimeout || 1000;
```
josdejong, line 258: Can you move the `1000` into a constant and use that constant twice?

src/worker.js
```diff
@@ -121,16 +130,34 @@ worker.cleanupAndExit = function(code) {
   var _exit = function() {
     worker.exit(code);
   }
-
-  if(!worker.terminationHandler) {
-    return _exit();
+  var _abort = function() {
+    worker.abortListeners = [];
```
josdejong, line 134: For a little bit of optimization, can you enclose this in an if statement that checks whether there are listeners and only if so, create a new empty array?

josdejong (commented):
Thanks for the updates! I added a few inline comments.

About the WorkerHandler terminating the worker when receiving a `cancel` or `timeout` error: that is a very good point. Do you have ideas on how to solve this? Indeed both the worker side and the WorkerHandler side both need to know whether there is an abort listener in place or not, and adjust their behavior accordingly.

joshLong145:
> Thanks for the updates! I added a few inline comments.
> 
> About the WorkerHandler terminating the worker when receiving a `cancel` or `timeout` error: that is a very good point. Do you have ideas on how to solve this? Indeed both the worker side and the WorkerHandler side both need to know whether there is an abort listener in place or not, and adjust their behavior accordingly.

Sorry for the delayed response. After giving it some thought I see a possible flow for communication which will allow for proper tracking of handlers based on if a worker can be reused after the `OnAbort` handlers are triggered

```mermaid
sequenceDiagram
    participant Pool
    participant WorkerHandler
    participant Worker
    Pool->>WorkerHandler: TimeoutError/CamcelationError occures, move task with rosolver to `tracking` queue. Send a message to the worker to run cleanup with the task id
    WorkerHandler ->> Worker: Worker recieves message, execute abort handlers.
    Worker ->> WorkerHandler: Send the result of abort handler execution to the worker handler with the task id sent
    WorkerHandler ->> Pool: Check the task id for a tracking and if present either resolve or reject the resolver promise based on the data sent in the message from the worker. Cleanup the task context
    
```
With the above model, the `resolver` Promise created when `exec` is first called on the `WorkerHandler` will either `resolve` or `reject` based on the result of the message sent back from the `onAbort` listener execution.
Which will be looked up from a `tracking` queue. The pool can now have a concept of tasks which need to be tracked for potential future cleanup. Since Cleanup operations are a parallel operation which requires resource tracking a second queue seems like the most obvious way of managing the resource.


The other idea, although much more involved is to rewrite how items are processed on the producer. Instead of items only being processed in a single promise chain with a recursive call. We could use something like [p-queue](https://www.npmjs.com/package/p-queue) to handle assigning tasks to workers and managing `WorkerHandlers` if tasks cause a worker to be terminated.

josdejong:
Ow nice I didn't know that you can draw a sequenceDiagram straight in GitHub issues 😎. 

I'm not sure whether a queue is really needed since a WorkerHandler and a Worker only process a single task at a time, but at least both need to get the information on whether to abort or not.

I think we could do something like this (it's close to your diagram I think): 

1. In case of a timeout error or termination request, the `WorkerHandler` sends a message to the `Worker` to abort. Then we have three possible situations:
    1. The `Worker` responds with "abort succesful". After that the `WorkerHandler` can terminate the `Worker`
    2. The `Worker` responds with "no abort callback". After that the `WorkerHandler` can terminate the `Worker`
    3. The `Worker` does not respond (is blocked by heavy CPU for example). After a limited timeout, the `WorkerHandler` will terminate the `Worker` anyway.
2. In case of a cancel error, the `WorkerHandler` sends a message to the `Worker` to abort. Then we have three possible situations:
    1. The `Worker` responds with "abort succesful". Then we're done.
    2. The `Worker` responds with "no abort callback". After that the `WorkerHandler` can terminate the `Worker`
    3. The `Worker` does not respond (is blocked by heavy CPU for example). After a limited timeout, the `WorkerHandler` will terminate the `Worker` anyway.

What do you think?

joshLong145:
> Ow nice I didn't know that you can draw a sequenceDiagram straight in GitHub issues 😎.
> 
> I'm not sure whether a queue is really needed since a WorkerHandler and a Worker only process a single task at a time, but at least both need to get the information on whether to abort or not.
> 
> I think we could do something like this (it's close to your diagram I think):
> 
> 1. In case of a timeout error or termination request, the `WorkerHandler` sends a message to the `Worker` to abort. Then we have three possible situations:
>    
>    1. The `Worker` responds with "abort succesful". After that the `WorkerHandler` can terminate the `Worker`
>    2. The `Worker` responds with "no abort callback". After that the `WorkerHandler` can terminate the `Worker`
>    3. The `Worker` does not respond (is blocked by heavy CPU for example). After a limited timeout, the `WorkerHandler` will terminate the `Worker` anyway.
> 2. In case of a cancel error, the `WorkerHandler` sends a message to the `Worker` to abort. Then we have three possible situations:
>    
>    1. The `Worker` responds with "abort succesful". Then we're done.
>    2. The `Worker` responds with "no abort callback". After that the `WorkerHandler` can terminate the `Worker`
>    3. The `Worker` does not respond (is blocked by heavy CPU for example). After a limited timeout, the `WorkerHandler` will terminate the `Worker` anyway.
> 
> What do you think?

Your outline makes sense and aligns with what the diagram outlines but with better definitions of possible execution results. I think you have mapped out most of the remaining details. However, I think we might want an explicit case for when there are `abort callbacks` but they have thrown an error.  While this will result in the same outcome as if the worker was stuck in some CPU bound operation we should handle it explicitly.

josdejong:
Thanks. That makes sense indeed, the abort handler can throw an error too.

src/worker.js
joshLong145: Are we able to make this static? since at time of definition `worker` will not have `emit` defined? I think it might be fine since `worker` is the root object and will be referenced passed.

joshLong145:
Hey @josdejong 

I think I have made good progress and was able to implement the feature set we have discussed above where 

1) We are able to prevent the termination of a worker in the pool if the `abortListeners` resolve and no errors occur within scope of a listener
2) We can timeout the execution of `abortListeners` such that if they are too long running / never leave scope we can short circuit and terminate the worker.
3) If there is an error within the listener we can handle the `rejection` and terminate the worker.

I have added tests for both `timeout` and `cancel` behaviors to confirm both have the desired behaviors.

I still have to update the docs / examples but I think it is ready for review.

src/worker.js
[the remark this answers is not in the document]
joshLong145: I ended up being able to use the `all` implementation within our `Promise` implementation. I refactored to use it as adding new Promise methods within this pr felt out of scope and turned out not to be necessary.

test/Pool.test.js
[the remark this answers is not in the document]
joshLong145: @josdejong Do we wish to allow for `emit` to extend to tasks which are being `tracked` as this PR now introduces it as a concept internally for cleanup support. If we do allow lookup in this map based on the `id` in the payload then we could support event emitting during task cleanup.

josdejong:
Thanks for the updates, that sounds good. I'll look into it after the holidays (sorry for the delay)

test/Pool.test.js
```diff
@@ -1315,6 +1315,196 @@ describe('Pool', function () {
...
+    
+    let task = pool.exec('asyncTimeout', [],  {});
+
+    // Wrap in a new promise which waits 50ms
+    // in order to allow the function executing in the
+    // worker to 
+    return new Promise(function(resolve) {
+      setTimeout(function() {
+        resolve();
+      }, 50);
```
josdejong, line 1358: Maybe overkill for this unit test, but instead of a "blind" waiting of 50 ms, you could let the worker emit an event as soon as it started executing the task, and use that as trigger.

see: [Events](https://github.com/josdejong/workerpool?tab=readme-ov-file#events)

test/Pool.test.js
```diff
@@ -1315,6 +1315,196 @@ describe('Pool', function () {
...
+    var pool = createPool({
+      onCreateWorker: () => {
+        workerCount += 1;
+      },
+      maxWorkers: 1,
+    });
+    function asyncTimeout() {
+      var me = this;
+      return new Promise(function () {
+        let timeout = setTimeout(function() {
```
josdejong, line 1422: how can we know for sure that this timeout is not triggered during the test? Maybe let it throw an exception or print "THIS SHOULD NOT BE PRINTED" or something?

Or: create a flag `let aborted = false`, and set that to true inside the `me.worker.addAbortListener`, and at the end verify that it is indeed true?

test/Pool.test.js
```diff
@@ -1315,6 +1315,196 @@ describe('Pool', function () {
...
+  
+    return pool.exec('asyncAbortHandlerNeverResolves', [])
+    .timeout(1000)
+    .catch(function (err) {
+      assert(err instanceof Promise.TimeoutError);
+
+      var stats = pool.stats();
+      assert(stats.busyWorkers === 1);
+    }).always(function() {
+      return pool.exec(add, [1, 2]).then(function() {
```
josdejong, line 1472: Instead, shouldn't we check here that `stats.totalWorkers` is zero?

test/Pool.test.js
```diff
@@ -1315,6 +1315,196 @@ describe('Pool', function () {
...
+    var pool = createPool(__dirname + '/workers/cleanup-abort.js', {
+      maxWorkers: 1,
+      onCreateWorker: () => {
+        workerCount += 1;
+      }
+    });
+    
+    return pool.exec('asyncTimeout', [],  {
+    })
+    .timeout(200)
```
josdejong, line 1329: Indentation missing here I think.

(We should set up prettier for this project to automatically format the code 😅 )

src/WorkerHandler.js
```diff
@@ -399,17 +418,46 @@ WorkerHandler.prototype.exec = function(method, params, resolver, options) {
...
+          }); 
+      });
+      
+      me.worker.send({
+        id,
+        method: CLEANUP_METHOD_ID 
+      });
+
+      /**
+       * Sets a timeout to reject the cleanup operation if the message sent to the worker
```
josdejong, line 445: Thanks for this explanation.

Can you also write documentation about this new feature in the README.md?

test/Pool.test.js
```diff
@@ -1315,6 +1315,196 @@ describe('Pool', function () {
...
+        resolve();
+      }, 50);
+    }).then(function() { 
+      return task.cancel()
+      .catch(function (err) {
+        assert(err instanceof Promise.TimeoutError);
+        var stats = pool.stats();
+        assert(stats.busyWorkers === 1);
+      }).always(function() {
+        return pool.exec(add, [1, 2]).then(function() {
```
josdejong, line 1501: I'm really a bit confused about all the `pool.exec(add, [1, 2])` in the unit tests, I would love to see a more explicit solution like with `stats().totalWorkers`. Or maybe I'm totally misunderstanding the reason for `pool.exec(add, [1, 2])` in the unit tests 😅 .

josdejong (changes requested):
This feature was _way_ more complex than I had anticipated at first. But the implementation that you have created now looks really solid, impressive job Bean!

I made a couple of inline comments. It looks like a lot but it is nothing big. Can you have a look at those?

test/Pool.test.js
```diff
@@ -1315,6 +1315,196 @@ describe('Pool', function () {
...
+    
+    return pool.exec('asyncTimeout', [],  {
+    })
+    .timeout(200)
+    .catch(function (err) {
+      assert(err instanceof Promise.TimeoutError);
+      let stats = pool.stats();
+      assert(stats.activeTasks === 0);
+    }).then(function() { 
+      return pool.exec(add, [1, 2])
```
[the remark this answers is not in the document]
joshLong145: What invoking another function to run in the pool gives us that checking on the `pool stats` doesn't is confirmation that the pool is not creating a new worker to service the request and instead uses the `worker` which should already be created and ready for another task as once the `then` is invoked the task has been moved to a `tracking` queue. By invoking a method such as `add` it lets us be sure the cleanup operations work as expected.

test/workers/cleanup-abort.js
```diff
@@ -0,0 +1,46 @@
...
+function asyncAbortHandlerNeverResolves() {
+  var me = this;
+  return new Promise((resolve) => {
+    let timeout = setTimeout(() => {
+        resolve();
+    }, 5000); 
+    me.worker.addAbortListener(function () {
+      clearTimeout(timeout);
+      return new Promise((res) => {
+        setTimeout(() => {
```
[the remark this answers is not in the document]
joshLong145: I thought about this but I thought the comically high timeout resolve was easier to understand.

test/Pool.test.js
joshLong145: Good call. updating to reject with an error instance instead of resolving with `void`

test/Pool.test.js
joshLong145: Wouldn't the worker count be `1` here since we have finished the task and the `add` task did not cause the worker to terminate? here are the stats from the pool after `add` finishes

```
activeTasks = 0
busyWorkers = 0
idleWorkers = 1
pendingTasks = 0
totalWorkers = 1
```

This makes sense to me as the total number of workers is `1` and the total number of idle workers is `1` so the worker can pick up the next task queued. This is the behavior I think we want correct?

test/Pool.test.js
joshLong145: I feel using calling another such as `add` is required to fully test this new feature as the tests are trying to prove that that a new worker is never created after the first, and due to the design of `pool` we create new workers at time of need. So by calling another method, such as `add` we can assert on this behavior and show that the pool resource provisioning is working as expected. We can still add calls to `pool.stat()` and assert on the worker counts. But I feel that the invocation of another method is beneficial to assert the pool is managing `worker` instances in the ways we expect.

test/Pool.test.js
joshLong145: I tried using an event, however after some digging into the implementation shown [here](https://github.com/josdejong/workerpool/blob/cdb5861133fbce9c911545308871b81ecb6a36cd/src/WorkerHandler.js#L273). I found that is not possible without refactoring to handle `tacked` tasks since the `task` queue is checked and if a task is not found we simply continue processing tasks.

test/Pool.test.js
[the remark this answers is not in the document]
josdejong: ah, that makes sense. Would it add much complexity to support `emit` from within the abort handler? If so I think it's not worth it.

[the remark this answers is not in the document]
josdejong: o wait I see I'm replying on an outdated comment, please ignore...

test/workers/cleanup-abort.js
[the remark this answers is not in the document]
josdejong: Fine with me 👍

test/Pool.test.js
josdejong: Ok let's not put effort in that just for the sake of the unit tests, thanks for checking this out.

test/Pool.test.js
josdejong: Sorry, yes I mean testing `stats.totalWorkers === 1`, not zero.

test/Pool.test.js
josdejong: That is true, it is a good extra test indeed calling a method right after it. I think we want to ensure two things happen:

1. After abort, a worker is _not_ terminated, so workerCount remains 1 and should not become 0.
2. When executing another task (like `add`) after aborting, the existing worker must be used. I think for this it makes sense to set `maxWorkers: 2` (so there is room for workerpool to mistakenly create a 2nd worker) and ensure the workerCount is still 1 after executing `add`.

I think by only testing `workerCount === 1` _after_ executing `add`, we will miss when aborting would not work correctly and would terminate the worker. I think it is essential to verify the worker count before and after `add`.

joshLong145: Good points!

joshLong145:
@josdejong 
Made the updates. Going to start on Docs/example.

joshLong145:
@josdejong 
After running the test suite with `--detect-leaks` I have found the test which is causing the `Pool` test suite timeout
`should clear all workers upon explicit termination`

Will look into it now that I have an area of focus.

josdejong:
Thanks for the updates. The PR is almost ready now! Remaining things:

1. Fix the `recieve` typo
2. Can you please a check to the unit tests testing `assert.strictEqual(stats.totalWorkers, 1)` after cancelation and before calling `pool.exec(add, [1, 2])`? Or is that already taken care of in a different way?
3. Fix the unit test timeout issue

Please let me know if you need help debugging this timeout.

joshLong145:
> 2\. assert.strictEqual(stats.totalWorkers, 1



> Thanks for the updates. The PR is almost ready now! Remaining things:
> 
> 1. Fix the `recieve` typo
> 2. Can you please a check to the unit tests testing `assert.strictEqual(stats.totalWorkers, 1)` after cancelation and before calling `pool.exec(add, [1, 2])`? Or is that already taken care of in a different way?
> 3. Fix the unit test timeout issue
> 
> Please let me know if you need help debugging this timeout.

Sorry about that, I had stashed some of the changes I made to the tests. They are now pushed.


In terms of debugging help. I realized that the flag I mentioned above was not in fact useful. However I have found some tests which do not fully terminate which are linked [here](https://github.com/joshLong145/workerpool/blob/64e3b9ba775c5871c20f4dbe57311c56548ef956/test/Pool.test.js#L497)

I have found that in some cases the following call to resolve a promise never completes: https://github.com/joshLong145/workerpool/blob/feat/abort-listener/src/WorkerHandler.js#L588

I am not sure why this does not resolve at this point. but fairly confident that it prevents the `after` hook (here](https://github.com/joshLong145/workerpool/blob/64e3b9ba775c5871c20f4dbe57311c56548ef956/test/Pool.test.js#L19) from resolving. It seems to be an infinite recursion bug, as `resolve` never leaves scope it seems from breakpointing and seeing the callstack loop.

josdejong:
Sounds good, thanks!

I've done some debugging. The timeout error occurs in `Pool.test.js`. I've removed all as much tests as possible to create a minimal version of `Pool.test.js` that still reproduces the issue:

```js
var assert = require('assert');
var {Promise} = require('../src/Promise');
var Pool = require('../src/Pool');

describe('Pool', function () {
  it('should timeout a task', function () {
    var pool = new Pool({maxWorkers: 10})

    function forever() {
      return new Promise(resolve => setTimeout(resolve, Infinity))
    }

    return pool.exec(forever)
      .timeout(50)
      .finally(() => {
        console.log('terminating...')
        return pool.terminate()
      }).finally(() => {
        console.log('terminated')
      });
  });
});
``` 

Here, the `pool.terminate()` promise never resolves. I haven't figured out why, but CPU stays at zero, so there is not an infinite loop but for some reason terminate is waiting for some callback that never comes. Any thoughts?

joshLong145:
> Sounds good, thanks!
> 
> I've done some debugging. The timeout error occurs in `Pool.test.js`. I've removed all as much tests as possible to create a minimal version of `Pool.test.js` that still reproduces the issue:
> 
> ```js
> var assert = require('assert');
> var {Promise} = require('../src/Promise');
> var Pool = require('../src/Pool');
> 
> describe('Pool', function () {
>   it('should timeout a task', function () {
>     var pool = new Pool({maxWorkers: 10})
> 
>     function forever() {
>       return new Promise(resolve => setTimeout(resolve, Infinity))
>     }
> 
>     return pool.exec(forever)
>       .timeout(50)
>       .finally(() => {
>         console.log('terminating...')
>         return pool.terminate()
>       }).finally(() => {
>         console.log('terminated')
>       });
>   });
> });
> ```
> 
> Here, the `pool.terminate()` promise never resolves. I haven't figured out why, but CPU stays at zero, so there is not an infinite loop but for some reason terminate is waiting for some callback that never comes. Any thoughts?

I came to the same conclusion in the comment above. Working on figuring our cause. I notice that if you modify the `createPool` hook to the following the worker does terminate

```js
  function createPool(script, options) {
    var pool = new Pool(script, options);

    after(() => {
      return pool.terminate(true, 1);
    });

    return pool;
  }
```

Thinking now its an issue with how promises are chained in `WorkerHandler.exec` but still debugging.

joshLong145:
So With commit c68cc9b I have been able to get `Pool should timeout a task` to complete tear down, along with all other tests within `pool`. So it looks like scheduling a task which settles the promise from within  the `catch` of that promise caused the resolver to never leave scope.

 However,  if I run them in serial. However, when run in parallel with enough parallelism there seems to be one or more tests which do not fully tear down. I think it is a similar issue to what was observed with `should timeout a task` as there is something keeping the final promise from resolving. When attaching the chrome debugger to the test harness I do always see `two workers` left once all the tests conclude. But I am a bit stumped as

josdejong:
> So With commit [c68cc9b](https://github.com/josdejong/workerpool/commit/c68cc9b7ab2f57bddabeed9bcd6368111971abef) I have been able to get `Pool should timeout a task` to complete tear down

This is a tough problem 😅 . When running the `should timeout a task` on it's own in the latest version of the PR I still get a 10s timeout. 

To discover where the issue originates, it may help to put a `console.log(...)` at all important stages to see which parts of the code are executed and in which order (and compare that with a test that _does_ work as expected).

joshLong145:
> > So With commit [c68cc9b](https://github.com/josdejong/workerpool/commit/c68cc9b7ab2f57bddabeed9bcd6368111971abef) I have been able to get `Pool should timeout a task` to complete tear down
> 
> This is a tough problem 😅 . When running the `should timeout a task` on it's own in the latest version of the PR I still get a 10s timeout.


Interesting, How are you running the test and what nodes version are you using?

josdejong:
I checked out the latest version of the PR, running on node 20.17.0, Windows 11. I exect though that that doesn't make a difference. When running `npm test`, it ends with the following error:

```
1) "after all" hook in "{root}":
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; 
       if returning a Promise, ensure it resolves.
``` 

This is the same error that we see in the tests running on Github Actions (running on Linux, node 18, 20, 22): https://github.com/josdejong/workerpool/actions/runs/10802599625/job/29964991022?pr=448

When only running this one test `it.only('should timeout a task', function () { ... })` I get the exact same error message.

joshLong145:
@josdejong 
After reading [mocha hook docs](https://mochajs.org/#migrating-tests-to-use-root-hook-plugins) I was curious on the use of `after` since it is a global cleanup hook that runs once per test suite. I have modified the `createPool` hook to `afterEach` as we make a new `pool` in many test cases so to me the use of the `after` hook should be `afterEach` as we want a new pool on each test *individually* and for each pool to clean up after the test completes. 


When I make this refactor all tests pass and terminate successfully, the result is that each test suite concludes operation and mocha is able to terminate the process. Test now pass and complete in CI as well. I do not have access to a machine running `Windows` at this time. Could you re run the suite and see if you get the same result?

josdejong:
Thanks for your updates. Running the latest version of this PR on my Windows machine indeed works and all tests pass. 

However, I don't think that this is an issue related to Mocha but some other race condition introduced in the code of this PR. When reducing `Pool.test.js` to a single test that I posted at https://github.com/josdejong/workerpool/pull/448#issuecomment-2333598456, the unit test fails. When extracting the code from this test in a Node.js script and running it without Mocha, it also still fails (never prints "terminated"). 

I'm afraid we really have to dig deeper to figure out where this issue is coming from before merging this PR.

joshLong145:
> Thanks for your updates. Running the latest version of this PR on my Windows machine indeed works and all tests pass.
> 
> However, I don't think that this is an issue related to Mocha but some other race condition introduced in the code of this PR. When reducing `Pool.test.js` to a single test that I posted at [#448 (comment)](https://github.com/josdejong/workerpool/pull/448#issuecomment-2333598456), the unit test fails. When extracting the code from this test in a Node.js script and running it without Mocha, it also still fails (never prints "terminated").
> 
> I'm afraid we really have to dig deeper to figure out where this issue is coming from before merging this PR.

Thanks for testing. I can reproduce the [#448 (comment)](https://github.com/josdejong/workerpool/pull/448#issuecomment-2333598456) on my machine. I still think that the `after` hook should instead be `afterEach` but this isolated test failure needs addressing. 

going to keep digging, in the mean time going to flag this as a draft PR until I can diagnose root cause.

josdejong:
Thanks! If I have some time I'll also do some debugging to see if I can find anything.

joshLong145:
@josdejong 


After tweaking the `finally` implementation to not return itself on either `resolve` or `reject` of the internal promise wrapper we now seem to have a passing test per https://github.com/josdejong/workerpool/pull/448#issuecomment-2333598456. With the change I just pushed the isolated test case now passes for me.

josdejong:
Wow, good find!!! 👏 

I've done some more digging, and the issue originates in the method `Promise.finally` not handling promises returned by the callback correctly, like in our failing unit test `.finally(() => pool.terminate());`. So it looks like this issue is _not_ introduced in this PR but we discovered an existing bug 😅 .

Just removing `.then(res)` from `.finally` is not a good solution though: it is there for a reason, namely propagating the original promise chain (and value or errror). But there where no unit tests guarding this. I've worked out a PR to fix the issue with `Promise.finally` and add more unit tests: #470. Can you revert your change in `Promise.finally`? I'll merge this PR soon, and if you update your branch with the fix from #470 this issue will be solved.

src/Promise.js
```diff
@@ -224,7 +224,7 @@ Promise.prototype.finally = function (fn) {
   const final = function() {
     return new Promise(function (resolve) {
       resolve(fn())
-    }).then(res);
```
josdejong, line 227: Can you revert this change? (PR #470 will solve the issue)

src/WorkerHandler.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 
+const { EventEmitter } = require('node:events');
```
josdejong, line 3: Can you remove this redundant import?

src/worker.js
```diff
@@ -2,21 +2,49 @@
  * worker must be started as a child process or a web worker.
  * It listens for RPC messages from the parent process.
  */
+const { addAbortListener } = require('events');
```
josdejong, line 5: Can you remove this redundant import?

test/Pool.test.js
```diff
@@ -15,7 +15,7 @@ describe('Pool', function () {
   function createPool(script, options) {
     var pool = new Pool(script, options);
 
-    after(() => {
+    afterEach(() => {
```
josdejong, line 18: Can you change this back to `after(...)`? I think using `afterEach` will call `pool.terminate` on every pool after every test, but we only need to call it once per pool.

josdejong (changes requested):
A few small remarks, otherwise I think we're good to go. 

We should first merge #470 though.

josdejong:
Ok I've now merge #470. Can you update the PR to the latest version of the `master` branch with updated `Promise.finally`?

test/Pool.test.js
joshLong145: Don't we want to terminate the pool after every test? since the instance is created in scope if individual test cases. If we run clean up using `after` then all registered hooks run after the entire `Pool` `describe` block finishes executing.  I could be misunderstanding the purpose of the hook, but it seems like we want to clean up after each test as per how each `pool` instance is scoped to a test case.

joshLong145:
> Ok I've now merge #470. Can you update the PR to the latest version of the `master` branch with updated `Promise.finally`?

Merged in,I was thinking the return in the initially `finally` would suffice for returning the promise, but you are correct in that preserving the chain is still important. Good catch!

test/Pool.test.js
josdejong: If you create a global `count` in `Pool.test.js` and adjust the `afterEach` to increment and log the count, you'll see that `count` grows to 1801, whereas we only call `createPool()` 66 times. That is because we keep terminating all pools that have been created since the start after every test again and again.

```js
// global
let count = 0

// ...

afterEach(() => {
  count++
  console.log('count', count)
  return pool.terminate();
});
```

One solution is using `after` instead of `afterEach`. That has one downside: all created pools stay in memory until the end, so that costs more memory. 

An more neat solution is to change `createPool` to create a pool and then add that to a global list with created pools. Then, we create a single `afterEach` (outside of `createPool`) which calls `pool.terminate()` on every pool in the global array and then clear the array. That way we will terminate each created pool only once, and we terminate all created pools after every test. Shall we go for that? Ie:

```js
  // Creating pool with this function ensures that the pool is terminated
  // at the end of the test, which avoid hanging the test suite if terminate()
  // hadn't been called for some reason
  let createdPools = []
  function createPool(script, options) {
    const pool = new Pool(script, options);
    createdPools.push(pool);
    return pool;
  }

  afterEach(async () => {
    while (createdPools.length > 0) {
      await createdPools.shift().terminate();
    }
  });
```

josdejong:
When changing `after` into `afterEach` right now, we get a timeout error again. Somehow we still get some race condition when creating multiple pools. I've done some debugging and found the following minimal code demonstrating the issue:

```js
// file: terminatest.js
const { Promise } = require('./src/Promise');
const Pool = require('./src/Pool');

run()

async function run() {
  console.log('start')

  const pool1 = new Pool();
  await pool1.exec(sleep, [100])
    .timeout(50)
    .catch(err => { console.error(String(err)) })

  const pool2 = new Pool();
  const result = await pool2.exec(add, [3, 4])
  console.log({ result })

  await sleep(1000)
  
  console.log('pool1 workers:', pool1.workers.length) // <-- is 1 but should be 0

  await pool1.terminate() // <-- never finishes
  await pool2.terminate()

  console.log('done')
}

function sleep(delay) {
  return new Promise((resolve) => setTimeout(resolve, delay))
}

function add(a, b) {
  return a + b
}
```

I did some debugging but didn't yet found the cause of this. Some observations:

1. This only happens when creating 2 pools. This suggests that somehow these pools are not fully isolated from each other. I did look if we mess up the context somewhere (`this` and `me`) but didn't see anything.
2. The issue is gone when reverting the following piece of code to what it is in the master branch:
    `// on cancellation, force the worker to terminate`
    https://github.com/josdejong/workerpool/pull/448/files#diff-7e092d00b94d598e9df27965dd8a8157654881fcd56f5735e873d45c32c47146R418-R470 
    (though this effectively disables the abort listener and all these tests fail).
3. By adding console.log statements, I noticed that the `.timeout(50)` causes the worker of `pool1` to be terminated twice, I think that should be one. And the callback of `WorkerHandler.prototype.terminate` is not invoked at all. When adding the following lines to `WorkerHandler.prototype.terminate`, the issue disappears:

    ```js
    WorkerHandler.prototype.terminate = function (force, callback) {
      if (this.terminated && typeof callback === 'function') {
        return callback()
      }

      // ...
    ```

But (3) is not the right solution, we should probably look into why the terminate callback is not invoked in `terminateAndNotify` and/or why the workerhandler of `pool1` is not fully terminated when the `.timeout(50)` is hit. Maybe we miss calling `cleanup()` somewhere in `WorkerHandler.prototype.terminate`?

src/worker.js
```diff
@@ -134,12 +171,67 @@ worker.cleanupAndExit = function(code) {
...
+    });
+
+
+    return Promise.all([
+      settlePromise,
+      timeoutPromise
+    ]);
+  }
+  // if there are no listeners just reject in a promise and let the worker cleanup start
+  return new Promise(function(_resolve, reject) { reject(); });
```
josdejong, line 208: Can you explain this to me? It feels odd triggering a rejection when there is no error occurring, and it is also a bit odd to leave err `undefined`, i.e. should be something like `reject(new Error('...'))`

test/Pool.test.js
joshLong145: Ah, yes this makes sense without a aggergate like an `Array` to track pools seems like a good solution here. I think it is out of scope of this PR But will be a nice follow up.

src/worker.js
joshLong145: Good spot and yes we should be throwing an instance of `Error` here. I have updated this and also removed the `worker.kill()` invocation of the `tryCleanup` method as it will be handled through the `workerHandler` `cleanup` function.


With these changes I now get the following output from your isolated test implementation


```
start
TimeoutError: Promise timed out after 50 ms
{ result: 7 }
pool1 workers: 0
done
```


I have also reverted the `afterEach` hook back to `after` see https://github.com/josdejong/workerpool/pull/448#discussion_r1786896894

test/Pool.test.js
josdejong: 👍 I'll update the logic separately from this PR

josdejong:
I tested the issue I mentioned [here](https://github.com/josdejong/workerpool/pull/448#issuecomment-2390936044) again and and all works smooth now 🎉 .  I guess the underlying bug is fixed via https://github.com/josdejong/workerpool/pull/448/commits/35a08d2524b63874e5fd671257ddd5caaaa3ad04?

joshLong145:
> I tested the issue I mentioned [here](https://github.com/josdejong/workerpool/pull/448#issuecomment-2390936044) again and and all works smooth now 🎉 . I guess the underlying bug is fixed via [35a08d2](https://github.com/josdejong/workerpool/commit/35a08d2524b63874e5fd671257ddd5caaaa3ad04)?

Yes 😄  
Explained in this [comment](https://github.com/josdejong/workerpool/pull/448#discussion_r1786935211)

src/worker.js
joshLong145: I wanted to move this logic into `tryCleanup` where we send the message to the main thread the cleanup result. With the case of `worker.abortListeners.length === 1` returning a promise tracking the result of the handlers or a timeout promise. not returning a promise in the case of `worker.abortListeners.length === 0` diverges the return value of the function depending on the length of the `abortListeners` array.

joshLong145: I did a bit of refactoring and we could refactor `tryCleanup` in the following way

```js
worker.cleanup = function(requestId) {
  var _exit = function() {
    worker.exit();
  }

  var _abort = function() {
    if (!worker.abortListeners.length) {
      worker.abortListeners = [];
    }
  }

  if (worker.abortListeners.length) {
    const promises = worker.abortListeners.map(listener => listener());
    let timerId;
    const timeoutPromise = new Promise((_resolve, reject) => {
      timerId = setTimeout(function () { 
        reject(new Error('Timeout occured waiting for abort handler, killing worker'));
      }, worker.abortListenerTimeout);
    });

    // Once a promise settles we need to clear the timeout to prevet fulfulling the promise twice 
    const settlePromise = Promise.all(promises).then(function() {
      clearTimeout(timerId);
      _abort();
    }, function() {
      clearTimeout(timerId);
      _exit();
    });


    return Promise.all([
      settlePromise,
      timeoutPromise
    ]).then(function() {
      worker.send({
        id: requestId,
        method: CLEANUP_METHOD_ID,
        error: null,
      });
    }, function(err) {
      worker.send({
        id: requestId,
        method: CLEANUP_METHOD_ID,
        error: err ? convertError(err) : null,
      });
    });
  }

  worker.send({
    id: requestId,
    method: CLEANUP_METHOD_ID,
    error: convertError(new Error('Worker terminating')),
  });
  return;
}
```

I am not loving this as there is one branch where the method returns a `Promise` and another where the function is retuning `void` but we still need to `await` the call to `cleanup` as if `abortListeners` has a handler we must await it. 

I like the previous implementation better as it allows a promise to be return in either branch which to me is a better implementation, but curious what your thoughts are.

josdejong: Hm, yeah 🤔.

I think the confusion on my side originates from the cleanup function having three different cases, but returning only two cases: 

1. there where abort listener(s) and they succesfully aborted, so no need to terminate the worker
2. there where abort listener(s) but something went wrong, so we need to terminate the worker 
3. there where no abort listeners, so we need to terminate the worker

Now, returning two cases works fine because in case (2) and (3) we need to take the same action: terminate the worker. We could implement something to report back these three different cases to the WorkerHandler to make it explicit, or we can leave it as it is right now and clarify with some extra comments.

Idea:

- Rename the function `tryCleanup` to `cleanup`
- Rename function `cleanupAndExit` to `terminateAndExit` to prevent any confusion with `cleanup`.
- Move all logic inside `cleanup` (like your last proposal), and add a comment explaining the three cases. Having it contained in one function with a bit of extra explanation may make it easier to understand and refactor in the future.
- Change the function `cleanup` to _always_ return `Promise<void>`
- Change function `cleanupAndExit` to return `Promise<void>` too (it contains a promise but no return right now)
- Change the error message "Cleanup failed, exiting worker" into "Worker terminating", that is more clear to me.
- Change `worker.on('message', function (request) {...}` into `worker.on('message', async function (request) { ...}`, remove the two `return` cases, replace them with `await`.
- Extra: reduce nesting in function `cleanup` by exiting early in case of `worker.abortListeners.length === 0`, moving this `if` to the top of the function and inverting it. That will also make `_abort` redundant I think, and you could inline `_exit` since it is a one liner which is only used once.

josdejong:
Awesome! I'm really happy that this last vague bug is resolved now 😅 !

Let's try to improve `tryCleanup` a bit, see my latest comment: https://github.com/josdejong/workerpool/pull/448#discussion_r1789908153, and after that we can merge the PR I think.

src/worker.js
joshLong145: I think we ultimately will need to implement some piece of context to correspond to the result of task execution. If we do this then there are many solutions to cleanup of worker handlers which are found to be in some terminal execution state and need to be killed and restarted. 

I was hesitant to implement this with this pr as I was unsure if we wanted to go down this route. I think for now we should go with your outline as changes to how workers are terminated could have adverse effects on up stream projects.

I think the approach you outline as the initial solution makes sense but I have a question regarding

> Change worker.on('message', function (request) {...} into worker.on('message', async function (request) { ...}, remove the two return cases, replace them with await

Would we effect up stream projects with the use of `async / await` in code that is released? If so we might want to keep with the current option of returning a promise. So we might need to keep the pattern of returning a promise in both cases of either `successful` or `error` the handler execution may result in.

joshLong145: > The main reason to make the callback async is to prevent vague (future) issues when there are multiple async actions inside the callback that we would expect to run after each other. Using await in the code makes it clear that an async function is being executed.



Can't we achieve this with the current pattern of returning the `Promise` from `cleanup` in the `worker` message handler? 

I think if the user wants to implement multiple async actions from within the callback then they may declare the callback function to be `async`. WDYT?

josdejong: Thanks for the updates, I think the function `cleanup` now contains enough explanation for my futher me to understand what's going on 😉. 

> Can't we achieve this with the current pattern of returning the `Promise` from `cleanup` in the `worker` message handler?

Yes, that is also fine with me 👍. I like the `async` and return `Promise<void>` slightly better but if you prefer this approach I'm fine with that.

Last remark: I think the following line is now sort of redundant, and can be reduced to just `return`, or returning a resolved promise. Then, the function consistently returns `Promise<void>` in all three cases. Does that make sense?

```js
return new Promise(function(reject) { reject(new Error("Worker terminating"));
```
The custom `Promise` of workerpool doesn't have a `Promse.resolve`, so this will look like `return new Promise(resolve => resolve());`

josdejong (approved):

josdejong:
I think we're there, clicking the Merge button now 🎉 . Thanks a lot for your patience and perseverance Josh, this was a long journey with quite some bumps along the road!

josdejong:
Published now in `v9.2.0`, thanks again!

joshLong145:
> Published now in `v9.2.0`, thanks again!

No problem, was great working on this :)

GwiYeong:
@joshLong145 Hello, It's a great feature for me. thanks for your working. 
I'm trying to test it with the example you gave.
But i noticed that when i output stdout like `console.log` inside `addAbortListener` function, parent doesn't know about it.
Is it possible to emit stdout to parent from within `addAbortListener` as well?
this is simple example
parents.js
```js
const main = async () => {
    const cleanedUpTask = pool.exec('asyncTimeout', [], {
        on: function (payload) {
            if (payload.stdout) {
                console.log(`[Worker] Out: ${payload.stdout.trim()}`)
            }
        }
    }).timeout(1_000).catch((err) => {
        console.log("task timeout");
        console.log("timeout occured: ", err.message);
        console.log("worker count ", workerCount);
        return pool.exec(add, [1, 2]).then((sum) => {
            console.log('add result', sum);
            console.log("worker count: ", workerCount);
        });
    });
    await cleanedUpTask;
...
}
main()
```
worker.js
```js
function asyncTimeout() {
  var me = this;
  console.log('trigger AsyncTimeout')
  return new Promise(function (resolve) {
    let timeout = setTimeout(() => {
        resolve();
    }, 5000);

    me.worker.addAbortListener(async function () {
        console.log('clearTimeout')
        clearTimeout(timeout);
        resolve();
    });
  });
}
```
output is 
```
New worker created
[Worker] Out: trigger AsyncTimeout
task timeout
timeout occured:  Promise timed out after 1000 ms
worker count  1
add result 3
worker count:  1
```

joshLong145:
@GwiYeong 
I originally thought adding support for events from workers would not be easily supported for abort handlers as I would be extending the task life cycle. However, it turns out it was easier than I originally thought.

I have raised a PR here: #478

joshLong145:
@GwiYeong
Thanks for looking into this, could you open a separate issue for investigation?

GwiYeong:
@joshLong145 I created https://github.com/josdejong/workerpool/issues/479. Thanks!

Issue #447, opened 2024-06-04
Is it possible to use workerpool with Webpack 5 for browsers?


im7mortal:
I couldn't find in the documentation if it can be used with Webpack 5 for Web Workers.

So in [the official documentation](https://webpack.js.org/guides/web-workers/) it's told:

```javascript
new Worker(new URL('./worker.ts', import.meta.url));

new Worker(
  /* webpackChunkName: "foo-worker" */ new URL('./worker.js', import.meta.url)
);
```

Webpack 5 will recognize that the TypeScript file inside the URL within the Worker must be compiled separately. It will then substitute the path string with a unique hash name of the separately generated file.

However, since workerpool doesn't use the new Worker(new URL("path")) pattern, Webpack 5 cannot compile it separately and perform the substitution.

## Possible workaround for now

I use workerpool for my personal project, so I can experiment with unconventional solutions. I know that inside the worker we can actually get the path of the file with:

```javascript
self.addEventListener('message', function(event) {
  console.log('Worker script path:', self.location.href);
});
```

Therefore, I am considering initializing the worker with a function that will retrieve the href and then initialize workerpool with it.

josdejong:
Have you tried the webpack5 example?

https://github.com/josdejong/workerpool/tree/master/examples/webpack5

im7mortal:
@josdejong Thank you! 
It worked! 

I am thinking if it would be possible to contribute to docs so it would be easier to find. I will close issue after

josdejong:
Good to hear. 

A PR improving the docs would be welcome :)

josdejong:
I'll close this issue since the question is answered. A PR improving the docs is still welcome of course.

PR #453, opened 2024-06-20
docs: clarify webpack5 setup with workerpool (#447)

im7mortal:
I added mentions about React or webpack5 builds in the web worker sections


josdejong:
Thanks Petr!

PR #455, opened 2024-06-20
fix: transfer from worker fixed

LeonidPolukhin:


w1nklr (approved):

josdejong:
Thanks for debugging and fixing this issue Leonid.

Can you add a unit test too, to prevent future regressions?

w1nklr:
Hi @josdejong 
I'm afraid it's the same problem than https://github.com/josdejong/workerpool/pull/439#issuecomment-2034032764.

Again, if there is a way to do unit tests in browser env, I would be more than happy to (learn the way to do and) to implement the test ! But for now, we're clueless with @LeonidPolukhin ...

josdejong:
Ok makes sense. Let's not try to write an automated test then. 

Do you have a minimal browser example that I can test manually to verify the solution?

LeonidPolukhin:
> Ok makes sense. Let's not try to write an automated test then.
> 
> Do you have a minimal browser example that I can test manually to verify the solution?

I could add one more example in, for instance, examples/webpack5 or extend the existing one. Which variant would you prefer?

LeonidPolukhin:
> I'll check with @LeonidPolukhin
> 
> I saw there is an example with webpack5 (`examples/webpack5`). We might be able to enrich this example...

Extended the example in webpack 5. Worker creates an array of the requested size and transfers it back to the main thread using workerEmit. After that it returns boolean value indicating whether the array buffer is detached. With the current version of workerpool the example shows that arrays are cloned but with the fix applied they are transferred.

josdejong:
Thanks, I tried it out and indeed it works like expected 👍

josdejong:
The fix is published now in `v9.1.3`, thanks again!

PR #456, opened 2024-06-28
chore: update webpack5 example to illustrate worker feedback.

w1nklr:
As a followup of #439, here comes an example of how to use workerpool.workerEmit() in a browser.

New example (left: before computations, right: with computations):
image

Enhanced readme to warn that the used `workerpool` package can not be up-to-date and add instructions to install latest package.


w1nklr:
I wanted to add it to PR #455, but couldn't push it to Leonid's branch, and then it was too late :/

josdejong:
Thanks @w1nklr, this looks good! 

I think we should not expand this example further 😅, in general I try to give each example a focus and let it explain 1 thing else you don't see the trees from the forest.

One last idea: how about rewriting the example to for rexample React? All the manual management of `document.createElement(...)` etc inside a JSX file looks a bit odd 😁 (I think that was originally copied from a plain JS example).

josdejong:
ping @w1nklr

w1nklr:
Sorry, I'm quite busy these times...

Just checking I understood it correctly: you want to introduce a dev dependency on react and rewrite this (and in the end all) example in react. Correct ?

josdejong:
Yes indeed. But only if you feel like it. 

We can leave it as it is for now, it was just an idea for further improvement.

w1nklr:
I just updated this one example to react.
Let me know if this is what you want.

But build with node (22) fails at test level :(

josdejong:
Thanks, I'll review the code soon.

The node 22 issue is not caused by this PR itself. It seems that there is something changed in the Github Actions or the latest Ubuntu version that it uses, it doesn't support using globals like `rollup` inside a project for some reason. I have something similar in an other project. No clue how to fix it though.

josdejong:
The node 22 issue was introduced in `v22.5.0` and fixed in `v22.5.1` (see https://github.com/nodejs/node/pull/53934)

w1nklr:
Great.
The checks are green now !

w1nklr:
You're welcome.
Thank you for the ping, this gave me the push I needed ;)

PR #481, opened 2024-12-05
Add oncreated worker

GwiYeong:
I wanted to add a cpu limit for the worker process created when utilizing workerpool in process mode.
However, the current callback doesn't know the pid of the created worker.

This PR adds a callback to get the information of the created worker after the worker is created.
I added a callback called `onCreatedWorker`, which is passed the currently created WorkerHandler directly.
If it's a process worker, I can get the created pid via `worker.worker.pid`.
The pid can then be used to limit resources, such as cpu limit.

For cpulimit worker, I add an example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,7 @@ The following options are available:
   - `workerOpts: Object`: the `workerOpts` option of this pool
   - `script: string`: the `script` option of this pool
     Optionally, this callback can return an object containing one or more of the above properties. The provided properties will be used to override the Pool properties for the worker being created.
+- `onCreatedWorker: Function`. A callback that is called whenever a worker is created. For example, it can be used to add a limit, such as a cpu limit, for each created worker. 
 - `onTerminateWorker: Function`. A callback that is called whenever a worker is being terminated. It can be used to release resources that might have been allocated for this specific worker. The callback is passed as argument an object as described for `onCreateWorker`, with each property sets with the value for the worker being terminated.
 - `emitStdStreams: boolean`. For `process` or `thread` worker type. If `true`, the worker will emit `stdout` and `stderr` events instead of passing it through to the parent streams. Default value is `false`.
 
--- a/src/Pool.js
+++ b/src/Pool.js
@@ -51,6 +51,8 @@ function Pool(script, options) {
   /** @readonly */
   this.onCreateWorker = options.onCreateWorker || (() => null);
   /** @readonly */
+  this.onCreatedWorker = options.onCreatedWorker || (() => null);
+  /** @readonly */
   this.onTerminateWorker = options.onTerminateWorker || (() => null);
 
   /** @readonly */
@@ -421,7 +423,7 @@ Pool.prototype._createWorkerHandler = function () {
     script: this.script
   }) || {};
 
-  return new WorkerHandler(overriddenParams.script || this.script, {
+  const worker = new WorkerHandler(overriddenParams.script || this.script, {
     forkArgs: overriddenParams.forkArgs || this.forkArgs,
     forkOpts: overriddenParams.forkOpts || this.forkOpts,
     workerOpts: overriddenParams.workerOpts || this.workerOpts,
@@ -431,6 +433,8 @@ Pool.prototype._createWorkerHandler = function () {
     workerTerminateTimeout: this.workerTerminateTimeout,
     emitStdStreams: this.emitStdStreams,
   });
+  this.onCreatedWorker(worker)
+  return worker;
 }
 
 /**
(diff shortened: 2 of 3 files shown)


josdejong:
Thanks @GwiYeong. Thinking aloud here: can we already achieve this already by using events? Would it be possible to let a worker send it's pid to the main process via `workerpool.workerEmit`?

GwiYeong:
@josdejong Thank you for responding.
I just wonder that is there any way to receive messages from a worker that are passed to the `workerpool.workerEmit` function without calling `pool.exec`?
In case of cpu limit, it should be called only once per worker, and if the worker is terminated and recreated, it needs to be executed again. 
Is there a way to send a message to `workerpool.workerEmit` as soon as the worker is created, depending on the life cycle of the worker, and have the main process receive and handle it?
If it is possible, can you give me some example code to achieve it?

josdejong:
You can indeed only send/receive `workerpool.workerEmit` whilst executing a task. But a new worker will only be created when there is a task to execute. You can do do something like this:

```js
// myWorker.js
var workerpool = require('workerpool')

var initialized = false

function wrapperSendPidOnce(fn) {
  return (...args) => {
    if (!initialized) {
      initialized = true

      workerpool.workerEmit({ type: 'created', pid: process.pid })
    }

    return fn(...args)
  }
}

function add(a, b) {
  return a + b
}

function multiply(a, b) {
  return a * b
}

workerpool.worker({
  add: wrapperSendPidOnce(add),
  multiply: wrapperSendPidOnce(multiply)
})
```
And then use the worker like so:
```js
// main.js
var workerpool = require('workerpool')

var pool = workerpool.pool(__dirname + '/myWorker.js')

function execWithEventListener(method, args) {
  function handleEvent(event) {
    console.log('Event: ', event)
  }

  return pool.exec(method, args, { on: handleEvent })
}

async function run() {
  console.log(await execWithEventListener('add', [2, 3]))
  console.log(await execWithEventListener('add', [7, 2]))
}

run()
  .catch(function (err) {
    console.error(err)
  })
  .finally(function () {
    pool.terminate()
  })
```

Running this example wil output: 

```
Event:  { type: 'created', pid: 26312 }
5
9
```

Would that address your case?

GwiYeong:
@josdejong  yes. I understand what you are saying.
However, It would be easier if we had a lifecycle hook like `onCreatedWorker`, but it seems a bit complicated to determine whether it's the first time or not.
If a worker's functions should always be executed by calling
How about adding a function like `pool.broadcast` that triggers all the worker's functions and handles all workers emit?
If so, we could separate functions in the worker that return only the pid and for action.

joshLong145:
Maybe I am missing something, but there is already an event which is triggered when workers get created called `onCreateWorker` which is passed arguments containing the worker options and the script it is executing. for example
```js
{
  forkArgs: [],
  forkOpts: {},
  workerOpts: { type: "module" },
  workerThreadOpts: {},
  script: "blob:null/8c54e7b5-6cde-4bc8-9a0e-b3020ff4b694"
}
```
This event is currently called [here](https://github.com/josdejong/workerpool/blob/badc05b16d5f7a719e8f6c142779bb2743d5d183/src/Pool.js#L416) when a new worker gets created.
Could these event arguments be extended to include the `pid` in the case of a `process` worker? This event would then meet your use case and we would not need a new `onCreatedWorker` event type.

joshLong145:
This is an interesting idea, I agree that since there is an `onCreateWorker` an `onCreatedWorker` makes sense. As it extends the life cycle model and does not require explicit implementation within worker contexts to implement. However, if we are to add a `onCreatedWorker` perhaps the life cycle as a whole should be thought of and perhaps we should also add `onTerminatedWorker` to complete cases of pre and post termination / creation of workers.

 >  it seems a bit complicated to determine whether it's the first time or not
 What do you mean by first time?

GwiYeong:
@joshLong145 
> it seems a bit complicated to determine whether it's the first time or not
> What do you mean by first time?
I mean the solution that @josdejong gave to me achieve my goal, but it seems a bit complicated. 
`wrapperSendPidOnce` can always be executed only after the worker's function has been executed, so this is a slightly different timing than when the worker is created. 
If I want to continuously monitor a worker process after it is created, I can't do that until the worker's function is called once (I don't even know the worker's PID.). I always need to run the worker's function after it is created, and also I can't specify what worker to run, I can't determine the timing of when a newly created worker can be monitored.

josdejong (changes requested):
Thanks @GwiYeong, and sorry for the late reply. 

I've been thinking about it and you're right, the workaround that I posted is quite complicated. Also, adding support for an `onCreatedWorker` option is only a few lines of code, so not much to worry there.

I made a couple of inline comments, can you have a look at those?

etorresh:
@josdejong this would be useful. I can finish this up with the requested changes. Should I create a new PR?

PR #483, opened 2024-12-18
fix: :bug: process.exitCode instead of calling process.exit to allow cleanup

### LINKED ISSUE #429: `worker.exit` should set `process.exitCode` instead of calling `process.exit` to allow cleanup

pcace:
closes https://github.com/josdejong/workerpool/issues/429

see https://github.com/josdejong/workerpool/issues/429

The test sometimes failed, i guess because it now takes a bit longer to finish code and exit. after using 2000 it never failed on my machine. Hope thats ok.

### DIFF
--- a/src/worker.js
+++ b/src/worker.js
@@ -81,7 +81,10 @@ else if (typeof process !== 'undefined') {
     var parentPort  = WorkerThreads.parentPort;
     worker.send = parentPort.postMessage.bind(parentPort);
     worker.on = parentPort.on.bind(parentPort);
-    worker.exit = process.exit.bind(process);
+    worker.exit = function (code) {
+      process.exitCode = code;
+      parentPort.postMessage('__workerpool-terminate__');
+    };
   } else {
     worker.on = process.on.bind(process);
     // ignore transfer argument since it is not supported by process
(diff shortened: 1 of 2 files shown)


src/worker.js
```diff
@@ -81,7 +81,10 @@ else if (typeof process !== 'undefined') {
     var parentPort  = WorkerThreads.parentPort;
     worker.send = parentPort.postMessage.bind(parentPort);
     worker.on = parentPort.on.bind(parentPort);
-    worker.exit = process.exit.bind(process);
+    worker.exit = function (code) {
+      process.exitCode = code;
+      parentPort.postMessage('__workerpool-terminate__');
```
joshLong145, line 86: I am confused on why we are seeing a message to the `parent` for shutdown when this message typically sent from parent to child, not child to parent. The parent message handler does not have implementations to deal with this message type.

Perhaps I am missing something, but to me this is not going to achieve the goals of providing the exit code on shutdown.

josdejong: Yes, I have the same question indeed, this message is only send from the main process to the worker right now. 

I would expect something like:

```js
worker.exit = function (code) {
  process.exitCode = code
  process.exit()
}
```

josdejong:
Thanks for your PR @pcace, and sorry for the late reply. 

Two remarks:

- Can you have a look at the inline comment of Bean?
- How can we reproduce your issue, so we can test whether this PR actually solves the issue? Is it possible to write a unit test to validate the fix?

pcace:
Hi there, I am very sorry, but I am not using it anymore, so my testcase is gone. 
its totally ok if you close the pr.

Issue #479, opened 2024-12-03
Terminate worker even if abortListener resolved


GwiYeong:
While testing abortListener, I noticed something strange.
As I understood, if the abortListener is resolved normally (when the promise is resolved within the abortListenerTimeout), the Worker is not Terminated and can continue to execute other functions.
But if you add onTerminateWorker to the example code [here](https://github.com/josdejong/workerpool/blob/v9.2.0/examples/abort.js), you can see that the Worker is Terminated. Since the workerTerminateTimeout is 1 second, the Worker will be terminated after 1 second.
Because [this line](https://github.com/josdejong/workerpool/blob/v9.2.0/examples/abort.js#L24) was done before the Worker was terminated, it looks like the Worker was not recreated.
Is this a bug? Or am I misunderstanding something? Am I right that the worker is terminated even though I originally registered the abortListener?
I only added 4 lines for testing. [this file](https://github.com/josdejong/workerpool/blob/v9.2.0/examples/abort.js)

```js
const pool = workerpool.pool(__dirname + "/worker.js", {
    ...
    onCreateWorker: (args) => {
        console.log("New worker created");
        workerCount += 1;
    },
+    onTerminateWorker: (args) => {
+        console.log("Worker terminated");
+        workerCount -= 1;
+    }
});
const main = async () => {
    ...
    await cleanedUpTask;
+    await new Promise((resolve) => setTimeout(resolve, 10000)); //wait few seconds
    ...
}
```

joshLong145:
@GwiYeong @josdejong

After changing [abort listener execution](https://github.com/joshLong145/workerpool/blob/85de98559b32e0287a335e52e895517f7fc45193/src/worker.js#L234) to use a race implementation over an `all` implementation. We are now seeing abort handlers resolve if the promise each handler returns fuffils. My understanding of task resolution was not totally accurate. With the migration to worker termination being async from [PR-176](https://github.com/josdejong/workerpool/pull/176) we now remove the `processing` task on `reject` of the task promise. So, when the `processing` promise fuffils we have no busy workers as there are no tasks in process, assuming no other tasks are queued.

This goes against how I previously thought of task life cycle as when a timeout triggers we will have `0` active tasks once the promise rejects as we clean up the processing task.

We are not tracking the state of the `tracked` tasks in our essement of a worker so the worker will not be `busy` if a task is being tracked. which allows it to be chosen as the worker for the next task.

[...]

PR #484, opened 2024-12-18
Fix/abort listener execution

joshLong145:
PR for discussion on issue #479

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -24,13 +24,15 @@
  * @property {import('worker_threads').WorkerOptions} [workerThreadOpts] Object`. For `worker` worker type. An object passed to [worker_threads.options](https://nodejs.org/api/worker_threads.html#new-workerfilename-options).
  * @property {boolean} [emitStdStreams] Capture stdout and stderr from the worker and emit them via the `stdout` and `stderr` events. Not supported by the `web` worker type.
  * @property { (arg: WorkerArg) => WorkerArg | undefined } [onCreateWorker] A callback that is called whenever a worker is being created. It can be used to allocate resources for each worker for example. Optionally, this callback can return an object containing one or more of the `WorkerArg` properties. The provided properties will be used to override the Pool properties for the worker being created.
- * @property { (arg: WorkerArg) => void } [onTerminateWorker] A callback that is called whenever a worker is being terminated. It can be used to release resources that might have been allocated for this specific worker. The callback is passed as argument an object as described for `onCreateWorker`, with each property sets with the value for the worker being terminated.
+ * @property { (arg: WorkerArg) => void } [onTerminateWorker] A callback that is called whenever a worker is being terminated. It can be used to release resources that might have been allocated for this specific worker. The callback is passed as argument an object as described for `onCreateWorker`, with each property sets with the value for the worker being terminated. 
  */
 
 /**
  * @typedef {Object} ExecOptions
  * @property {(payload: any) => unknown} [on] An event listener, to handle events sent by the worker for this execution.
  * @property {Object[]} [transfer] A list of transferable objects to send to the worker. Not supported by `process` worker type. See ./examples/transferableObjects.js for usage.
+ * @property {(payload: {id: number,  error: Error | undefined, isTerminating: boolean}) => void } [onAbortResolution] An event listener triggered when whenever an abort operation concludes.
+ * @property {(payload: {id number, taskResolver: any) => void } [onAbortStart] an event listener triggered when a task throws a Timeout or Canceltion exception and cleanup is starting. 
  */
 
 /**
--- a/src/types.js
+++ b/src/types.js
@@ -32,7 +32,7 @@
  * @property {(payload: any) => unknown} [on] An event listener, to handle events sent by the worker for this execution.
  * @property {Object[]} [transfer] A list of transferable objects to send to the worker. Not supported by `process` worker type. See ./examples/transferableObjects.js for usage.
  * @property {(payload: {id: number,  error: Error | undefined, isTerminating: boolean}) => void } [onAbortResolution] An event listener triggered when whenever an abort operation concludes.
- * @property {(payload: {id number, taskResolver: any) => void } [onAbortStart] an event listener triggered when a task throws a Timeout or Canceltion exception and cleanup is starting. 
+ * @property {(payload: {id number, AbortResolver: import('./Promise.js').Promise<void>}) => void } [onAbortStart] an event listener triggered when a task throws a Timeout or Canceltion exception and cleanup is starting. 
  */
 
 /**
--- a/src/types.js
+++ b/src/types.js
@@ -27,12 +27,27 @@
  * @property { (arg: WorkerArg) => void } [onTerminateWorker] A callback that is called whenever a worker is being terminated. It can be used to release resources that might have been allocated for this specific worker. The callback is passed as argument an object as described for `onCreateWorker`, with each property sets with the value for the worker being terminated. 
  */
 
+
+/**
+  * @typedef {Object} AbortStartArgs
+  * @property {number} [id] identifier of the task which is starting its abort operation.
+  * @property {PromiseLike<void>} [abortResolver] PromiseLike Object which resolves or rejects when the abort operation concludes.
+  *
+  */
+
+/**
+  * @typedef {Object} AbortResolutionArgs
+  * @property {Error | undefined} [error] An error which occured during the abort operation. If an error did not occure the value will be `undefined`.
+  * @property {number} [id] identifier of the task.
+  * @property {boolean} [isTerminating] A flag which indicates the termination status of the worker which ececuted the task. 
+  */
+
 /**
  * @typedef {Object} ExecOptions
  * @property {(payload: any) => unknown} [on] An event listener, to handle events sent by the worker for this execution.
  * @property {Object[]} [transfer] A list of transferable objects to send to the worker. Not supported by `process` worker type. See ./examples/transferableObjects.js for usage.
- * @property {(payload: any) => void} [onAbortResolution] An event listener triggered when whenever an abort operation concludes.
- * @property {(payload: any) => void} [onAbortStart] An event listener triggered when a task throws a Timeout or Canceltion exception and cleanup is starting. 
+ * @property {(payload: any) => AbortResolutionArgs} [onAbortResolution] An event listener triggered when whenever an abort operation concludes.
+ * @property {(payload: AbortStartArgs) => void} [onAbortStart] An event listener triggered when a task throws a Timeout or Canceltion exception and cleanup is starting. 
  */
 
 /**
--- a/src/types.js
+++ b/src/types.js
@@ -31,7 +31,7 @@
 /**
   * @typedef {Object} AbortStartArgs
   * @property {number} [id] identifier of the task which is starting its abort operation.
-  * @property {PromiseLike<void>} [abortResolver] PromiseLike Object which resolves or rejects when the abort operation concludes.
+  * @property {PromiseLike<void>} [abortPromise] PromiseLike Object which resolves or rejects when the abort operation concludes.
   *
   */
 
--- a/src/types.js
+++ b/src/types.js
@@ -24,7 +24,7 @@
  * @property {import('worker_threads').WorkerOptions} [workerThreadOpts] Object`. For `worker` worker type. An object passed to [worker_threads.options](https://nodejs.org/api/worker_threads.html#new-workerfilename-options).
  * @property {boolean} [emitStdStreams] Capture stdout and stderr from the worker and emit them via the `stdout` and `stderr` events. Not supported by the `web` worker type.
  * @property { (arg: WorkerArg) => WorkerArg | undefined } [onCreateWorker] A callback that is called whenever a worker is being created. It can be used to allocate resources for each worker for example. Optionally, this callback can return an object containing one or more of the `WorkerArg` properties. The provided properties will be used to override the Pool properties for the worker being created.
- * @property { (arg: WorkerArg) => void } [onTerminateWorker] A callback that is called whenever a worker is being terminated. It can be used to release resources that might have been allocated for this specific worker. The callback is passed as argument an object as described for `onCreateWorker`, with each property sets with the value for the worker being terminated. 
+ * @property { (arg: WorkerArg) => void } [onTerminateWorker] A callback that is called whenever a worker is being terminated. It can be used to release resources that might have been allocated for this specific worker. The callback is passed as argument an object as described for `onCreateWorker`, with each property sets with the value for the worker being terminated.
  */
 
 
@@ -39,7 +39,7 @@
   * @typedef {Object} AbortResolutionArgs
   * @property {Error | undefined} [error] An error which occured during the abort operation. If an error did not occure the value will be `undefined`.
   * @property {number} [id] identifier of the task.
-  * @property {boolean} [isTerminating] A flag which indicates the termination status of the worker which ececuted the task. 
+  * @property {boolean} [isTerminating] A flag which indicates the termination status of the worker which ececuted the task.
   */
 
 /**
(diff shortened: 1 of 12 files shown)


src/types.js
```diff
@@ -27,10 +27,28 @@
...
+  * @property {boolean} [isTerminating] A flag which indicates the termination status of the worker which ececuted the task.
+  */
+
 /**
  * @typedef {Object} ExecOptions
  * @property {(payload: any) => unknown} [on] An event listener, to handle events sent by the worker for this execution.
  * @property {Object[]} [transfer] A list of transferable objects to send to the worker. Not supported by `process` worker type. See ./examples/transferableObjects.js for usage.
+ * @property {(payload: any) => AbortResolutionArgs} [onAbortResolution] An event listener triggered when whenever an abort operation concludes.
+ * @property {(payload: AbortStartArgs) => void} [onAbortStart] An event listener triggered when a task throws a Timeout or Canceltion exception and cleanup is starting.
+ * @property {{ promise: import('./Promise.js').Promise<any, Error>; resolve: Function; reject: Function; }} [abortResolver] Defered Promise which resolves or rejects when the abort operation for the task concludes.
```
joshLong145, line 51: I am torn on if this is a good mechanism for allowing callers access to the promise which controls the abort operation. In `AbortResolutionArgs` we provide the promise for tracking the abort operations. We are able to provide the promise from the `tracking` `deferred` promise created internally.

We also allow the  `deferred` promise for the abort operation to be provided through the `ExecOptions`. Since we need a `deferred` promise we cannot simply provide an instance of `Promise`. 

This means the promise for the abort operation is exposed as two different types. Which is not ideal, but I am unsure if there is a way to unify them without breaking backwards compatibility.

src/types.js
```diff
@@ -27,10 +27,28 @@
...
+  * @property {Error | undefined} [error] An error which occured during the abort operation. If an error did not occure the value will be `undefined`.
+  * @property {number} [id] identifier of the task.
+  * @property {boolean} [isTerminating] A flag which indicates the termination status of the worker which ececuted the task.
+  */
+
 /**
  * @typedef {Object} ExecOptions
  * @property {(payload: any) => unknown} [on] An event listener, to handle events sent by the worker for this execution.
  * @property {Object[]} [transfer] A list of transferable objects to send to the worker. Not supported by `process` worker type. See ./examples/transferableObjects.js for usage.
+ * @property {(payload: any) => AbortResolutionArgs} [onAbortResolution] An event listener triggered when whenever an abort operation concludes.
```
josdejong, line 49: Is there an actual use case for the `onAbortResolution` and `onAbortStart` callbacks? If there is no clear use case I would prefer to not implement it right now, to keep things simple.

src/types.js
josdejong: I'm not sure how `abortResolver` would work, isn't it duplicate with `onAbortResolution`?

test/Pool.test.js
```diff
@@ -1,17 +1,17 @@
-var assert = require('assert');
-var {Promise} = require('../src/Promise');
-var Pool = require('../src/Pool');
-var tryRequire = require('./utils').tryRequire
+var assert = require("assert");
```
josdejong, line 1: I see you have different coding style configured in your IDE 😅 . We should really setup a linter for the project, let me see if I can set up Biomejs (if so we can run it after merging this PR, otherwise we'll have a lot of merge conflicts). 

Is it possible to revert the coding style changes, so I can see which unit tests are added/updated?

josdejong (commented):
Thanks Josh! I've made some inline comment, can you have a look at those?

src/types.js
joshLong145: onAbortResolution is called after the abort handlers have executed  and passes the taskId and error if the abort handlers threw an error, and a flag to determine if the worker is shutting down or not. 

`abortResolver` is just allowing the user to provide the promise that is used as the `tracking` promise for a tasks abort operation.

The idea is to give control of the promise to the caller, so they do not need to get it from event args. but it can be misused easily.

src/worker.js
```diff
@@ -236,10 +236,10 @@ worker.cleanup = function(requestId) {
   // - Reject if one or more handlers reject
   // Upon one of the above cases a message will be sent to the handler with the result of the handler execution
   // which will either kill the worker if the result contains an error, or 
-  return Promise.all([
-    settlePromise,
-    timeoutPromise
-  ]).then(function() {
+  return new Promise(function (resolve, reject) {
```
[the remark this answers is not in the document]
joshLong145: We can, but I was trying to not use any implementation of `Promise` from outside of the repo.

src/types.js
joshLong145: The main usage is to give more context to the user on life cycle of the task. but `onAbortStart` does provide access the to promise.
If we are to remove these callbacks then we should keep `abortResolver` so the user knows when the abort operation concludes.

test/Pool.test.js
joshLong145: Yes sorry I switched Editors recently.

joshLong145: I reverted back to the upstream master and applied my changes to the `abort handler` test suite. there are a couple formatting changes that I'll clean up. but the tests where largely re written to wait until the abort operation concludes to assert on the pool status. Which makes it a larger diff.

src/types.js
josdejong: I doubt whether there is much need for `onAbortResolution`, `onAbortStart`, and `abortResolver`, so at this point I prefer to keep the API as simple and limited as possible.

I understand that it can be handy for unit testing, but to make the unit tests work I think it can be okish to use workarounds like using checking the stats after a `setTimeout(..., 0)` and things like that.

> If we are to remove these callbacks then we should keep `abortResolver` so the user knows when the abort operation concludes.

I may be misunderstanding, but I would think that moment is when the promise of the task at hand rejects with a timeout or cancel error? In my head, the abort functionality only has an API at the _worker_ side and not at the _pool_ side. The pool side can cancel or timeout a task. Then, internally, we either terminate the worker or abort the task (keeping the worker runnig). End result from the pool side is that the promise resolves with a cancel or timeout error.

src/worker.js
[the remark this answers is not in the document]
josdejong: Ah of course, nevermind.

josdejong:
Hm I think I accidentally removed one of the inline comments, so if you mis something it was me, sorry 😅.

Most important right now is to get on the same page regarding https://github.com/josdejong/workerpool/pull/484#discussion_r1979834510.

src/types.js
joshLong145: >I doubt whether there is much need for onAbortResolution, onAbortStart, and abortResolver, so at this point I prefer to keep the API as simple and limited as possible. I understand that it can be handy for unit testing, but to make the unit tests work I think it can be okish to use workarounds like using checking the stats after a setTimeout(..., 0) and things like that.

We don't need them,  as you said we can still test without them with `setTimeout`. I thought lifecycle hooks would be a useful addition to users but I see your point on not over complicating the api.


>I may be misunderstanding, but I would think that moment is when the promise of the task at hand rejects with a timeout or cancel error? In my head, the abort functionality only has an API at the worker side and not at the pool side. The pool side can cancel or timeout a task. Then, internally, we either terminate the worker or abort the task (keeping the worker runnig). End result from the pool side is that the promise resolves with a cancel or timeout error.

This is mostly true, The main exception here is if the worker's abort operation times out and we have to reject the task promise from the pool side.

joshLong145:
> Hm I think I accidentally removed one of the inline comments, so if you mis something it was me, sorry 😅.
> 
> Most important right now is to get on the same page regarding [#484 (comment)](https://github.com/josdejong/workerpool/pull/484#discussion_r1979834510).

Responded in thread, I can remove the `onAbortStart` and `onAbortResolver`. Ill refactor tests to work without them.

josdejong:
Thanks 👍

josdejong:
@joshLong145 is this PR still on your radar?

joshLong145:
Yes, very sorry. I found an issue with the promise chain that I have fixed but have issues with cases dealing with worker crashes and properly rejecting the task promise.

I think it has to do with life cycle and trying to kill a worker that has already been killed and never receiving a message back to exit the worker on the second attempt to shutdown.

Hopefully I can get through this quickly.

josdejong:
Thanks for the update! Please let me know if I can help debugging anything.

joshLong145:
Will do, I am leaving for a trip tomorrow and will be back late next week. I plan to pick back up when I return.

josdejong:
Sure! Have a good time this week.

joshLong145:
I have a minimal example which highlights the issue I'm dealing with

```js
// example meant to me run within the `/examples` directory
const workerpool = require("..");

function main() {
  const pool = workerpool.pool(__dirname + "/workers/cleanupAbort.js", {
    // maximum time to wait for worker to cleanup it's resources
    // on termination before forcefully stopping the worker

    onCreateWorker: function (args) {
      console.log("New worker created");
    },
    onTerminateWorker: function () {
      console.log("worker terminated");
    },
    maxWorkers: 1,
  });

  return pool.exec("asyncAbortHandlerNeverResolves", [], {})
      .timeout(100);
      .catch(function() {
           console.log("this will log before the worker terminates");
      });
   
  return task;
};

main();
```

What I think is happening is that the promise returned [here](https://github.com/josdejong/workerpool/blob/d3e707fd5bcea1d19467e104f70403548f671d7e/src/WorkerHandler.js#L433) Is not chained when `timeout` is invoked since the timeout is reimplemented within `pool.exec`

joshLong145:
@josdejong  Below is a minimal example of the problem I am stuck on

```js
const workerpool = require("..");

function add(a, b) {
  return a + b;
}

function main() {
  const pool = workerpool.pool(__dirname + "/workers/cleanupAbort.js", {
    // maximum time to wait for worker to cleanup it's resources
    // on termination before forcefully stopping the worker

    onCreateWorker: function (args) {
      console.log("New worker created");
    },
    onTerminateWorker: function () {
      console.log("worker terminated");
    },
    maxWorkers: 1,
  });

  console.log("Starting first task that will timeout after 100ms...");

  // First task demonstrates our improved timeout handling
  return pool.exec("asyncAbortHandlerNeverResolves", [], {})
       .timeout(100)
       .catch(function(err) {
            console.log("Will log before the worker terminates");
       });
 
  return task;
};

main();
```

The above examples shows that the promise returned from the `timeout` call is not chained with the promise handling the worker termination [here](https://github.com/josdejong/workerpool/blob/d3e707fd5bcea1d19467e104f70403548f671d7e/src/WorkerHandler.js#L433).

josdejong:
Thanks for the clear, sort example. I would indeed expect the worker to be terminated before the `catch` callback is executed.

I'm not sure if the [`originalTimeout`](https://github.com/josdejong/workerpool/blob/d3e707fd5bcea1d19467e104f70403548f671d7e/src/Pool.js#L149) structure can be the cause of this issue though, this construct only "wraps" the original timeout function with a new function, but still calls the original timeout as far as I can see.

joshLong145:
> Thanks for the clear, sort example. I would indeed expect the worker to be terminated before the `catch` callback is executed.
> 
> I'm not sure if the [`originalTimeout`](https://github.com/josdejong/workerpool/blob/d3e707fd5bcea1d19467e104f70403548f671d7e/src/Pool.js#L149) structure can be the cause of this issue though, this construct only "wraps" the original timeout function with a new function, but still calls the original timeout as far as I can see.

Would it change your opinion if I pointed out the promise which is the `resolver` promise is not the same promise as what is returned from the call to [worker.exec](https://github.com/josdejong/workerpool/blob/d3e707fd5bcea1d19467e104f70403548f671d7e/src/WorkerHandler.js#L433). the promise returned from the call to the worker handler is not the resolver promise created in `pool.exec` which is what is being modified by [`originalTimeout`](https://github.com/josdejong/workerpool/blob/d3e707fd5bcea1d19467e104f70403548f671d7e/src/Pool.js#L149).

From my testing this seems to be true and why we are not seeing the worker terminate before the catch of the promise is executed.

josdejong:
You may definitely be right! If so, we should find an alternative solution to this wrapper functionality to solve this issue, right?

joshLong145:
> You may definitely be right! If so, we should find an alternative solution to this wrapper functionality to solve this issue, right?

Yes, I have tried things within scope of `worker handler.exec` to try and preserve the promise chain. But I am now thinking it needs to be handled higher up in either `pool._next` or `pool.exec`. I have a hunch that because `pool._next` doesn't return a promise that is where the chain needs to be preserved. 

So if I am right, we need to find a way to chain the promise created in `pool._next` to the resolver created in `pool.exec`.

josdejong:
Yes indeed 🤔

joshLong145:
@josdejong 
I have gotten much closer to a working solution. I have gotten the timeout promise to wait for the worker cleanups before fulfilling. However I am still running into issues with situations like below

```js
      var pool = new Pool({});

      function forever() {
        while (1 > 0) {} // runs forever
      }

      var promise = pool.exec(forever)
          .timeout(100)
          .then(function (result) {
            assert('promise should never resolve');
          })
        //.catch(Promise.CancellationError, function (err) { // TODO: not yet supported
          .catch(function (err) {
            conole.log(err);
            assert(err instanceof Promise.TimeoutError);
            assert.equal(pool.workers.length, 0);
            done();
          });
```


Where for some reason the `catch` on the task promise is not executing. So there is some place where the chain is breaking. I have spent a good amount of time trying to debug this but I have run into walls. I do not think I will be able to fully resolve the issues with this feature in the near future. 

I think the best option right now is to revert the abortListener functionality and I will come back with a new PR once I have the kinks worked out.

joshLong145:
@josdejong 
I have committed everything I have done up to this point here.
The two tests currently failing
`should timeout a task`
`should cancel a task`

I have also updated `examples/abort.js` to show that functionality is working in the `happy path` where one task does abort successfully keeping the worker, and another does not and kills the worker.

For the two offending tests it seems to be an issue where rejecting in the final termination promise [here](https://github.com/joshLong145/workerpool/blob/bf641934189c591d66ae9c2be392d0a0a123ee06/src/WorkerHandler.js#L458)
does not cause the `catch` of the timeout promise to resolve when the pool termination timeout triggers in the tracking promise.

I added a catch to the should timeout a task tests and confirmed the `catch` of the promise is not being invoked but I am not sure why at this point at it seems specific to the test case but I'm not sure what is special about this case as other calls to `timeout` work as expected.


My gut is currently saying it's a problem with how the task resolver is created. because we now create a new promise chain in the `resolvers` catch block I am thinking the system is not able to properly track this promise. as the system was not designed to track promises all the way down the chain as you have already called out.
I don't think we need to modify the current `promise implementation` to fix this issue, I think the implementation of the task resolver needs to be refactored to allow for chaining and to have the ability to keep the entire chain in tact while elevating the `resolve, reject` for invocation outside the scope of the promise.


The current implementation of `Promise.defer()` may be what needs rethinking instead of the entire promise implementation. I may be wrong of course, but this seems like the issue as I am not formally chaining the promise in the changes I have made [here](https://github.com/joshLong145/workerpool/blob/bf641934189c591d66ae9c2be392d0a0a123ee06/src/Pool.js#L164) to the task execution. 


Happy to keep working on this with you! I just think we should revert what is currently in `master` while we work out the bugs.

josdejong:
Thanks for the update. 

Seen the high complexity and hard to debug issues, I indeed think we should do some more thorough refactoring. You may be right that only the task resolver needs refactoring and not the promise, I'm not sure but let's keep any option in mind.

> Happy to keep working on this with you! I just think we should revert what is currently in `master` while we work out the bugs.

We can do two things: (a) publish a breaking change where we remove the abort listener functionality, or (b) document the known edge cases with it. What do you think is best?

joshLong145:
> Seen the high complexity and hard to debug issues, I indeed think we should do some more thorough refactoring. You may be right that only the task resolver needs refactoring and not the promise, I'm not sure but let's keep any option in mind.

Agreed, I was trying to highlight what I think is the issue with the current implementation.

>We can do two things: (a) publish a breaking change where we remove the abort listener functionality, or (b) document the known edge cases with it. What do you think is best?

If it is okay that the worker will terminate after the task promise fulfills then I can submit an mr that just refactors the worker logic. With updates to docs to explain the behavior. If this isnt a desired solution then I think we have to go with publishing a breaking change.

josdejong:
> I do have some code ready if we want to fix the current implementation.

That sounds promising!

PR #507, opened 2025-06-23
Minimum fix for Abort Listener

joshLong145:
Implements the minimum fix needed for the `Abort Listeners` to properly execute within the worker context. The change now allows the abort listener callback to be respected for either cleaning up the worker on error, or letting it continue to be used within the pool. As discussed in https://github.com/josdejong/workerpool/pull/484#issuecomment-2962689032.

There are a few minor updates to the abort tests to confirm that the workers are terminated in scenarios where the abort listener is intended to clean up the worker. This helps to confirm the current implementation is performing cleanup as expected.

### DIFF
--- a/src/worker.js
+++ b/src/worker.js
@@ -235,11 +235,12 @@ worker.cleanup = function(requestId) {
   // - Reject if one or more handlers exceed the 'abortListenerTimeout' interval
   // - Reject if one or more handlers reject
   // Upon one of the above cases a message will be sent to the handler with the result of the handler execution
-  // which will either kill the worker if the result contains an error, or 
-  return Promise.all([
-    settlePromise,
-    timeoutPromise
-  ]).then(function() {
+  // which will either kill the worker if the result contains an error, or keep it in the pool if the result
+  // does not contain an error.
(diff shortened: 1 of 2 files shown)


joshLong145:
| This is good news! Thanks Josh. So does this minimal PR fix https://github.com/josdejong/workerpool/issues/479?
Yes it does. This now properly respond to the `WorkerHandler` to either kill the worker or tell the handler it has recovered.


| Looking at the fix, I see there is only one change: replacing a Promise.all(...) in worker.js. So basically you replaced having to resolve both cases with resolving one of the two. Can we use Promise.any(...) there?

I think we already talked about this implementation in the comment thread here: https://github.com/josdejong/workerpool/pull/484#discussion_r1981537575

| Can you also update the comment above it since the line "Resolve once all handlers resolve" isn't true anymore?

I think the comment is still true as the promise will either resolve or reject once the handlers all resolve / reject, or the timeout triggers. I can add the missing detail on the timeout which will resolve the promise and terminate the worker.

josdejong:
Your fix is now published in `v9.3.3`

PR #549, opened 2026-04-30
Treat worker as busy while cleaning up canceled task

### LINKED ISSUE #548: "Worker terminated" from canceling a task
We got a "Worker terminated" error today when a user started an HTTP request that depended on a workerpool worker, interrupted the request, then immediately retried.

```
Error: Worker terminated
    at WorkerHandler.terminate (/home/app/deploy/node_modules/workerpool/src/WorkerHandler.js:516:45)
    at WorkerHandler.terminateAndNotify (/home/app/deploy/node_modules/workerpool/src/WorkerHandler.js:617:8)
    at ? (/home/app/deploy/node_modules/workerpool/src/WorkerHandler.js:456:26)
    at ? (/home/app/deploy/node_modules/workerpool/src/Promise.js:179:17)
    at ? (/home/app/deploy/node_modules/workerpool/src/Promise.js:109:7)
    at Array.forEach (<anonymous>)
    at _reject (/home/app/deploy/node_modules/workerpool/src/Promise.js:108:13)
    at Object.reject (/app/app/deploy/node_modules/workerpool/src/Promise.js:164:5)
    at Timeout._onTimeout (/home/app/deploy/node_modules/workerpool/src/WorkerHandler.js:484:36)
    at listOnTimeout (node:internal/timers:605:17)
```

The second task was started within the first task's `workerTerminateTimeout` and was apparently killed by it.

The following analysis is from Claude Code. I've done my best to review it and take responsibility for any mistakes.

## Summary

When a task is cancelled (via `Promise.cancel()`) or times out, `WorkerHandler` schedules a forced termination of the worker after `workerTerminateTimeout` (default 1000ms) to give the worker a chance to clean up gracefully. **During that grace window, the pool can dispatch a new, unrelated task onto the same worker.** When the timer fires (or the worker is otherwise force-terminated as part of the cancellation cleanup), every entry in `processing` is rejected with `Error: Worker terminated` — including the unrelated task that was assigned during the window.

The result: cancelling task A can cause task B, submitted milliseconds later, to fail with `Worker terminated` even though B has nothing to do with A's cancellation.

## Reproduction

```js
const workerpool = require('workerpool');

const pool = workerpool.pool({ maxWorkers: 1, workerTerminateTimeout: 100 });

const longTask = pool.exec(() => { while (true) {} }); // CPU-bound, can't cancel cooperatively
longTask.catch(() => {}); // ignore the cancellation rejection

setTimeout(() => {
  longTask.cancel();
  // Submit an unrelated task immediately. With maxWorkers: 1, the only worker
  // is currently in cleanup-after-cancel. It is reported by `busy()` as idle,
  // so the pool dispatches `add` onto it.
  pool.exec((a, b) => a + b, [3, 4])
    .then(r => console.log('result:', r))                  // never logs
    .catch(err => console.log('FAIL:', err.message));      // FAIL: Worker terminated
}, 50);
```

Output:
```
FAIL: Worker terminated
```

## Root cause

A worker can be torn down by two paths, but `WorkerHandler.busy()` only knows about one of them:

1. **`WorkerHandler.terminate()`** sets `this.cleaning = true` (`WorkerHandler.js:583`). `busy()` correctly reports the worker as busy and `Pool._getWorker()` skips it.
2. **`Promise.cancel()` / timeout in `exec`** is caught in the `resolver.promise.catch` block at `WorkerHandler.js:433` and adds the task to `tracking` with a scheduled `terminateAndNotify(true)` after `workerTerminateTimeout`. `busy()` does **not** consider `tracking`:

   ```js
   WorkerHandler.prototype.busy = function () {
     return this.cleaning || Object.keys(this.processing).length > 0;
   };
   ```

   So `Pool._getWorker()` (`Pool.js:273`) sees the worker as available and dispatches the next queued task. When the cleanup timer fires, `terminate(true)` rejects every entry in `processing` with `Error: Worker terminated` — sweeping up the newly-assigned task as collateral damage.

## Proposed fix

`busy()` should also report the worker as busy while it has entries in `tracking`:

```js
WorkerHandler.prototype.busy = function () {
  return this.cleaning
    || Object.keys(this.processing).length > 0
    || Object.keys(this.tracking).length > 0;
};
```

Once the worker responds to the cleanup message, the tracking entry is removed at `WorkerHandler.js:324` and the worker becomes idle again — so workers that successfully run their abort listener stay in the pool, just as they do today. Workers whose cleanup times out (i.e. the `workerTerminateTimeout` setTimeout fires) get force-terminated and removed from the pool — also as today. The only behavior change is that **newly arriving tasks are queued through the cleanup window instead of being dispatched onto a worker that's about to be killed**.

joshkel:
In case it helps, here's a draft PR to fix #548, created with assistance from Claude Code. I'm uncertain of your AI contributions policy - feel free to close this if you wish.

Fixes #548

Tests (via Claude Code):

- **New regression test** in `Pool.test.js`: `should not assign new tasks to a worker whose previous task was cancelled`. 
- **Two updated tests** in the `abort handler` group: the `… dedicated worker with Timeout` and `… dedicated worker with Cancellation` tests previously asserted `busyWorkers === 0` immediately after the cancel/timeout `.catch` fires — i.e., during the cleanup window. The corresponding **inline-worker** variants of those same tests already assert `busyWorkers === 1` at that point, so the dedicated-worker variants appear to have been written inconsistently. Updated them to match the inline variants. The post-`pool.exec(...)` assertions, which verify the worker was not replaced, are unchanged and still pass.

### DIFF
--- a/test/Pool.test.js
+++ b/test/Pool.test.js
@@ -685,6 +685,41 @@ describe('Pool', function () {
 
     // TODO: test whether a task in the queue can be neatly cancelled
 
+  it('should not assign new tasks to a worker whose previous task was cancelled', function (done) {
+    // Regression test: when a task is cancelled, WorkerHandler moves it to
+    // `tracking` and schedules a force-terminate after `workerTerminateTimeout`.
+    // During that window the worker must not pick up new tasks — otherwise
+    // the forced termination would reject them with "Worker terminated"
+    // even though they had nothing to do with the original cancellation.
+    var pool = createPool({maxWorkers: 1, workerTerminateTimeout: 50});
+
+    function forever() {
+      while (1 > 0) {} // runs forever (so cancel must force-terminate)
+    }
+
+    function add(a, b) {
+      return a + b;
+    }
+
+    var p1 = pool.exec(forever).catch(function (err) {
+      assert(err instanceof Promise.CancellationError);
+    });
+
+    // Cancel after the worker has actually started running `forever`.
+    setTimeout(function () {
+      p1.cancel();
+
+      // Queue a follow-up task immediately to ensure that it queues until
+      // the worker finishes cleaning up the cancelled task.
+      pool.exec(add, [3, 4])
+        .then(function (result) {
+          assert.strictEqual(result, 7);
+          done();
+        })
+        .catch(done);
+    }, 25);
+  });
+
   it('should timeout a task', function () {
     var pool = createPool({maxWorkers: 10});
 
@@ -1576,10 +1611,10 @@ describe('Pool', function () {
           let stats = pool.stats();
           assert.strictEqual(workerCount, 1);
           assert.strictEqual(stats.totalWorkers, 1);
-          assert.strictEqual(stats.idleWorkers, 1);
-          assert.strictEqual(stats.busyWorkers, 0);
-        }).then(function() { 
-          return pool.exec(add, [1, 2]) 
+          // Still busy - cleanup is in progress.
+          assert.strictEqual(stats.busyWorkers, 1);
+        }).then(function() {
+          return pool.exec(add, [1, 2])
         }).then(function() {
           var stats = pool.stats();
           assert.strictEqual(workerCount, 1);
@@ -1616,10 +1651,10 @@ describe('Pool', function () {
             let stats = pool.stats();
             assert.strictEqual(workerCount, 1);
             assert.strictEqual(stats.totalWorkers, 1);
-            assert.strictEqual(stats.idleWorkers, 1);
-            assert.strictEqual(stats.busyWorkers, 0);
-          }).then(function() { 
-            return pool.exec(add, [1, 2]) 
+            // Still busy - cleanup is in progress on the worker.
+            assert.strictEqual(stats.busyWorkers, 1);
+          }).then(function() {
+            return pool.exec(add, [1, 2])
           }).then(function() {
             var stats = pool.stats();
             assert.strictEqual(workerCount, 1);
--- a/src/WorkerHandler.js
+++ b/src/WorkerHandler.js
@@ -477,8 +477,8 @@ WorkerHandler.prototype.exec = function(method, params, resolver, options) {
         * We need this timeout in either case of a Timeout or Cancellation Error as if
         * the worker does not send a message we still need to give a window of time for a response.
         * 
-        * The workerTermniateTimeout is used here if this promise is rejected the worker cleanup
-        * operations will occure.
+        * The workerTerminateTimeout is used here if this promise is rejected the worker cleanup
+        * operations will occur.
       */
       me.tracking[id].timeoutId = setTimeout(function() {
           me.tracking[id].resolver.reject(error);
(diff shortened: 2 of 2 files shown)


josdejong:
Thanks Josh I hope to look into it next week.

test/Pool.test.js
```diff
@@ -685,6 +685,41 @@ describe('Pool', function () {
...
+      p1.cancel();
+
+      // Queue a follow-up task immediately to ensure that it queues until
+      // the worker finishes cleaning up the cancelled task.
+      pool.exec(add, [3, 4])
+        .then(function (result) {
+          assert.strictEqual(result, 7);
+          done();
+        })
+        .catch(done);
```
joshLong145, line 719: question: are we confident we are not just hitting the catch block and cleaning up? I think there's potential to swallow errors.

joshLong145:
@josdejong are you planning to review this? The bug being fixed here could easily be effecting others. If you feel comfortable I can do some testing and report back.

test/Pool.test.js
joshkel: `.catch(done)` was chosen to match the coding style of other tests. Replacing with something like

```js
      pool.exec(add, [3, 4])
        .then(function (result) {
          assert.strictEqual(result, 7);
        })
        .finally(done);
```

may be better. But, for consistency, I'd update the other tests as well (and do it in a separate PR).

No (meaningful) errors should be swallowed - `assert.strictEqual` failures are recorded even with `.catch(done)`. And it should be safe to assume that `pool.exec` itself succeeds because it's tested elsewhere.

joshLong145: This makes sense, we do consistenly use this pattern throughout the test suite. I did locally validate that the `catch` is never invoked through 

```js
         pool.exec(add, [3, 4])
           .then(function (result) {
             assert.strictEqual(result, 7);
             done();
           })
           .catch((e) => {
             assert.fail("task should not be rejected",);
           });
``` 

With this implementation the tests still pass and resolve, meaning we do not enter the `catch` and the test formally terminates. I also hooked some logging up to the `onCreateWorker` callback to verify that `2` workers are created. one at time of test start, and again on task cancelation.


With that said, I think merging this as is will be okay, the point on consistency makes sense. I'm fine with this merging as is.

josdejong: I agree that many of the tests could be improved, some are more verbose than needed. But indeed best to keep that for a separate refactoring PR and keep this PR focused on the actual bug fix.

josdejong:
Thanks Josh for this fix! And sorry for the delay, life happened.

This fix makes sense, thanks for working out this PR!

I personally do not use AI due to ethical concerns. As for PR's from others: most important to me is that someone takes ownership and responsibility for their own PR and actually understand the issue and the fix. I have to review more and more slop-PR's nowadays 😢 though I get better at recognizing them before spending too much time on them.

josdejong:
Fixed now in `v10.0.3`, thanks again Josh!

joshkel:
> Thanks Josh for this fix! And sorry for the delay, life happened.
>
> This fix makes sense, thanks for working out this PR!

No problem, and you're welcome. Thanks for following up, @josdejong.

PR #569, opened 2026-07-23
Use availableParallelism, not cpus().length when determining max workers

### LINKED ISSUE #568: Pool should use availableParallelism, not cpus, when determining maxWorkers
When determining the default `maxWorkers`, `Pool` leverages the number of CPUs reported by `os.cpus()` (via the environment module). This is an incorrect use of `os.cpus()` and the `os.availableParallelism()` function is provided for that purpose. The documentation for `os.cpus()` explicitly states:

> os.cpus().length should not be used to calculate the amount of parallelism available to an application. Use [os.availableParallelism()](https://nodejs.org/api/os.html#osavailableparallelism) for this purpose.

Why does this matter? In environments where the application is running inside of containers -- in my case an Ember.js build running in Kubernetes -- the `os.cpus().length` call returns the number of CPUs on the host machine; not the number of CPUs available to the container. This causes more parallelism in the worker pool than is actually efficient. In my use case, it also results in an out of memory error.

bdurette:
Resolves #568

The `os.cpus()` documentation explicitly recommends against using that function for determining the resources available to your process. This is, among other things, because it is not container-aware. Instead, we use `availableParallelism` when running outside the browser, which is the recommended alternative.

### DIFF
--- a/src/environment.js
+++ b/src/environment.js
@@ -24,7 +24,8 @@ module.exports.isMainThread = module.exports.platform === 'node'
   : typeof Window !== 'undefined';
 
 // determines the number of cpus available
+var os = require('os')
 module.exports.cpus = module.exports.platform === 'browser'
   ? self.navigator.hardwareConcurrency
-  : require('os').cpus().length;
+  : (((typeof os?.availableParallelism === "function") && os.availableParallelism()) || os.cpus().length);
 
--- a/src/environment.js
+++ b/src/environment.js
@@ -24,8 +24,7 @@ module.exports.isMainThread = module.exports.platform === 'node'
   : typeof Window !== 'undefined';
 
 // determines the number of cpus available
-var os = require('os')
 module.exports.cpus = module.exports.platform === 'browser'
   ? self.navigator.hardwareConcurrency
-  : (((typeof os?.availableParallelism === "function") && os.availableParallelism()) || os.cpus().length);
+  : require('os').availableParallelism();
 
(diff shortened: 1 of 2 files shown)


josdejong:
Thanks Brandon, good idea to use `availableParallelism()` instead of `cpus()`.

One thought: I see `availableParallelism` was introduced in Node.js v18, which has reached its end of life for quite some time now. Still, I think it would be neat to fallback to using `.cpus()` in case `.availableParallelism` isn't available, to keep things backward compatible. Only for the library code, not needed for the unit tests. What do you think?

bdurette:
> I think it would be neat to fallback to using `.cpus()` in case `.availableParallelism` isn't available, to keep things backward compatible. Only for the library code, not needed for the unit tests. What do you think?

I'm not opposed to that. Whatever you prefer is fine. I had implemented that in my [first commit](https://github.com/josdejong/workerpool/pull/569/changes/4990e94a20f058bc59049c30c71a5e2b3b090397) and then backed that out, because it added some complexity -- especially if I needed to test that fallback.

Take a look. If that looks good to you, we can roll back that latest commit that takes out the fallback easy enough.

josdejong:
Ah, I see your first commit. That is indeed what I mean, though I think I would write it like `os.availableParallelism ? os.availableParallelism() : os.cpus().length` but that is a matter of preference.

I indeed prefer keeping the backward compatibility, but I'm OK with not adding tests to validate that this fallback works (that would require mocking `os` or running on a very old node.js version I guess).

